PR #17, opened 2023-07-11
Support extensions using indexes reserved for internal organization

elliotmjackson:
Supplies a fallback mechanism for the deprecated extension index. Essentially we search for the index that we want, if its not found we look for the old one.

depends on bufbuild/protovalidate#43


Makefile
```diff
@@ -11,7 +11,7 @@ COPYRIGHT_YEARS := 2023
 LICENSE_IGNORE := -e internal/testdata/
 # Set to use a different compiler. For example, `GO=go1.18rc1 make test`.
 GO ?= go
-ARGS ?=
+ARGS ?=gc
```
elliotmjackson, line 14:
```suggestion
ARGS ?=
```

rodaine (approved):

PR #22, opened 2023-07-20
Add 'for_key' support

Alfus:
Updating conformance to v0.2.6


Alfus:
To distinquish violations from map keys (vs values)

rodaine (approved):

Issue #29, opened 2023-08-17
Make evaluator threadsafe without mutex


elliotmjackson:
Based on benchmarks, about 50% of CPU time is spent obtaining a read lock on this mutex. Ideally, this can be reworked to be thread-safe while minimizing the need to obtain a lock.

[example](https://github.com/bufbuild/protovalidate-go/pull/10)

emcfarlane:
Closing on #31 .

PR #31, opened 2023-08-18
Build validator copy cache on write

emcfarlane:
This was just fun to play with: Swap cache map on write. 😬 

Also tried a waitgroup, but that has worse performance than the r/w lock.
 
Idea for: https://github.com/bufbuild/protovalidate-go/issues/29

```
goos: darwin
goarch: arm64
pkg: github.com/bufbuild/protovalidate-go
                               │  current.txt  │               new.txt                │
                               │    sec/op     │    sec/op     vs base                │
Validator/ColdStart-8             210.9µ ± ∞ ¹   211.8µ ± ∞ ¹       ~ (p=1.000 n=1) ²
Validator/Lazy/Valid-8            136.1n ± ∞ ¹   104.3n ± ∞ ¹       ~ (p=1.000 n=1) ²
Validator/Lazy/Invalid-8          391.9n ± ∞ ¹   386.9n ± ∞ ¹       ~ (p=1.000 n=1) ²
Validator/Lazy/FailFast-8        136.90n ± ∞ ¹   97.42n ± ∞ ¹       ~ (p=1.000 n=1) ²
Validator/PreWarmed/Valid-8       101.1n ± ∞ ¹   102.8n ± ∞ ¹       ~ (p=1.000 n=1) ²
Validator/PreWarmed/Invalid-8     382.5n ± ∞ ¹   385.7n ± ∞ ¹       ~ (p=1.000 n=1) ²
Validator/PreWarmed/FailFast-8    93.68n ± ∞ ¹   95.46n ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                           476.5n         439.2n        -7.84%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                               │  current.txt  │                new.txt                │
                               │     B/op      │     B/op       vs base                │
Validator/ColdStart-8            421.4Ki ± ∞ ¹   421.4Ki ± ∞ ¹       ~ (p=1.000 n=1) ²
Validator/Lazy/Valid-8             0.000 ± ∞ ¹     0.000 ± ∞ ¹       ~ (p=1.000 n=1) ³
Validator/Lazy/Invalid-8           649.0 ± ∞ ¹     649.0 ± ∞ ¹       ~ (p=1.000 n=1) ³
Validator/Lazy/FailFast-8          168.0 ± ∞ ¹     168.0 ± ∞ ¹       ~ (p=1.000 n=1) ³
Validator/PreWarmed/Valid-8        0.000 ± ∞ ¹     0.000 ± ∞ ¹       ~ (p=1.000 n=1) ³
Validator/PreWarmed/Invalid-8      648.0 ± ∞ ¹     649.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
Validator/PreWarmed/FailFast-8     168.0 ± ∞ ¹     168.0 ± ∞ ¹       ~ (p=1.000 n=1) ³
geomean                                      ⁴                  +0.02%               ⁴
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
³ all samples are equal
⁴ summaries must be >0 to compute geomean

                               │ current.txt  │               new.txt                │
                               │  allocs/op   │  allocs/op    vs base                │
Validator/ColdStart-8            5.798k ± ∞ ¹   5.800k ± ∞ ¹       ~ (p=1.000 n=1) ²
Validator/Lazy/Valid-8            0.000 ± ∞ ¹    0.000 ± ∞ ¹       ~ (p=1.000 n=1) ³
Validator/Lazy/Invalid-8          15.00 ± ∞ ¹    15.00 ± ∞ ¹       ~ (p=1.000 n=1) ³
Validator/Lazy/FailFast-8         3.000 ± ∞ ¹    3.000 ± ∞ ¹       ~ (p=1.000 n=1) ³
Validator/PreWarmed/Valid-8       0.000 ± ∞ ¹    0.000 ± ∞ ¹       ~ (p=1.000 n=1) ³
Validator/PreWarmed/Invalid-8     15.00 ± ∞ ¹    15.00 ± ∞ ¹       ~ (p=1.000 n=1) ³
Validator/PreWarmed/FailFast-8    3.000 ± ∞ ¹    3.000 ± ∞ ¹       ~ (p=1.000 n=1) ³
geomean                                     ⁴                 +0.00%               ⁴
```

### DIFF
--- a/internal/evaluator/builder.go
+++ b/internal/evaluator/builder.go
@@ -15,7 +15,7 @@
 package evaluator
 
 import (
-	"sync"
+	"sync/atomic"
 
 	"buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go/buf/validate"
 	"github.com/bufbuild/protovalidate-go/internal/constraints"
@@ -29,12 +29,11 @@ import (
 // Builder is a build-through cache of message evaluators keyed off the provided
 // descriptor.
 type Builder struct {
-	mu          sync.Mutex
-	cache       map[protoreflect.MessageDescriptor]*message
+	cache       atomic.Pointer[map[protoreflect.MessageDescriptor]*message]
 	env         *cel.Env
 	constraints constraints.Cache
 	resolver    StandardConstraintResolver
-	Load        func(desc protoreflect.MessageDescriptor) MessageEvaluator
+	isLazy      bool
 }
 
 // NewBuilder initializes a new Builder.
@@ -45,30 +44,32 @@ func NewBuilder(
 	seedDesc ...protoreflect.MessageDescriptor,
 ) *Builder {
 	bldr := &Builder{
-		cache:       map[protoreflect.MessageDescriptor]*message{},
 		env:         env,
 		constraints: constraints.NewCache(),
 		resolver:    res,
+		isLazy:      !disableLazy,
 	}
 
-	if disableLazy {
-		bldr.Load = bldr.load
-	} else {
-		bldr.Load = bldr.loadOrBuild
-	}
-
+	cache := make(map[protoreflect.MessageDescriptor]*message, len(seedDesc))
 	for _, desc := range seedDesc {
-		bldr.build(desc)
+		bldr.build(desc, cache)
 	}
-
+	bldr.cache.Store(&cache)
 	return bldr
+}
+
+func (bldr *Builder) Load(desc protoreflect.MessageDescriptor) MessageEvaluator {
+	if bldr.isLazy {
+		return bldr.loadOrBuild(desc)
+	}
+	return bldr.load(desc)
 }
 
 // load returns a pre-cached MessageEvaluator for the given descriptor or, if
 // the descriptor is unknown, returns an evaluator that always resolves to a
 // errors.CompilationError.
 func (bldr *Builder) load(desc protoreflect.MessageDescriptor) MessageEvaluator {
-	if eval, ok := bldr.cache[desc]; ok {
+	if eval, ok := (*bldr.cache.Load())[desc]; ok {
 		return eval
 	}
 	return unknownMessage{desc: desc}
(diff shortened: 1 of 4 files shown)


internal/evaluator/builder.go
```diff
@@ -81,29 +79,36 @@ func (bldr *Builder) load(desc protoreflect.MessageDescriptor) MessageEvaluator
...
-	bldr.mtx.Lock()
-	defer bldr.mtx.Unlock()
-	return bldr.build(desc)
+	newCache := make(map[protoreflect.MessageDescriptor]*message, len(cache)+1)
+	for k, v := range cache {
+		newCache[k] = v
+	}
+	msgEval := bldr.build(desc, newCache)
+	bldr.cache.Store(&newCache)
+	return msgEval
```
rodaine, line 92: Would two concurrent calls to `loadOrBuild` potentially compete to get their evaluator into the cache with this?

```
A ---> Load Cache: C ---> Copy Cache: Ca ---> Cache New Eval ---> Store Cache: Ca \
                                                                                   >-- C?
B ---> Load Cache: C ---> Copy Cache: Cb ---> Cache New Eval ---> Store Cache: Cb /
```

Would using CAS help in this case?

internal/evaluator/builder.go
```diff
@@ -29,15 +29,11 @@ import (
 // Builder is a build-through cache of message evaluators keyed off the provided
 // descriptor.
 type Builder struct {
-	// TODO: (TCN-1708) based on benchmarks, about 50% of CPU time is spent obtaining a read
-	//  lock on this mutex. Ideally, this can be reworked to be thread-safe while
-	//  minimizing the need to obtain a lock.
-	mtx         sync.RWMutex
-	cache       map[protoreflect.MessageDescriptor]*message
+	cache       atomic.Pointer[map[protoreflect.MessageDescriptor]*message]
```
rodaine, line 32: since we're passing around the map into a bunch of functions, let's create a type for it (within internal):

```
type EvaluatorCache map[protoreflect.MessageDescriptor]*message
```

With the added benefit of being able to hang functionality off of it (like Clone) to make things a bit more ergonomic

internal/evaluator/builder.go
emcfarlane: Yep. A or B, one wins, next call might have to redo the work. Eventually the cache will be consistent. I've added back the lock to serialize writes. What do you mean by CAS?

internal/evaluator/builder.go
emcfarlane: Added the type!

emcfarlane: Renamed to `MessageCache` to avoid stutter lint.

internal/evaluator/builder.go
rodaine: [CompareAndSwap](https://pkg.go.dev/sync/atomic#Pointer.CompareAndSwap)

1. On load, hold onto the returned `prevPtr`
2. Make copy of the underlying Cache 
3. Build as before populating the evaluators into the Cache
4. Instead of calling `Store`, call `CompareAndSwap` using `prevPtr`
5. If successful, we're good to go. Done.
6. If it's not, it means there's a new value in the atomic.
7. Load again into `prevPtr`. 
8. Copy contents of the `prevPtr` into your copy of the cache.
9. Go to step 4.

Something like:

```go
newPtr := &newCache
for !bldr.cache.CompareAndSwap(prevPtr, newPtr) {
  prevPtr = bldr.cache.Load()
  for k, v := range *prevPtr {
    newCache[k] = v
  }
}
```

emcfarlane: Ah right! Thats neat. It would duplicate the work if both `A` and `B` are building the same message but avoid blocking on different message types. I'll try it.

rodaine: yeah unfortunately the only way to block concurrent duplicate work is with a mutex

emcfarlane: Benchmarking is the same between the two, could try add more to clarify but might be overkill. Let me know if anything is needed or want to look at something else

rodaine (approved):
LGTM, will re-review if you would like to extend the benchmarks to support multiple message types.

internal/evaluator/builder.go
emcfarlane: Okay, need to serialize writes to avoid a race on the `constaints.Cache`. I've got a benchmark with dynamic messages:
```go
type dynamicProto struct {
	proto.Message
	salt int32
}

func (d dynamicProto) ProtoReflect() protoreflect.Message {
	return dynamicMessage{d.Message.ProtoReflect(), d.salt}
}
//...
```
Each field is salted too. But I'll move it to an internal test.

emcfarlane: Race is on: 
- https://github.com/bufbuild/protovalidate-go/blob/e65d1395ab970830470fab8c928f4aadd3516c5d/internal/constraints/cache.go#L73
-  https://github.com/bufbuild/protovalidate-go/blob/e65d1395ab970830470fab8c928f4aadd3516c5d/internal/constraints/cache.go#L146

rodaine (approved):

PR #51, opened 2023-09-16
Bug: transitive field CEL expressions fail to resolve types during type checking

rodaine:
For custom field constraints with a type that's transitively included in the message (i.e., from a different proto file) were failing type checking in the CEL builder. This was the result of a subtle bug only adding types for the parent message but not for the field's message type itself.

A follow-up will be to add transitive tests upstream for conformance to share across the ecosystem.


internal/evaluator/builder.go
```diff
@@ -268,7 +268,7 @@ func (bldr *Builder) processFieldExpressions(
 	var opts []cel.EnvOption
 	if fieldDesc.Kind() == protoreflect.MessageKind {
 		opts = []cel.EnvOption{
-			cel.TypeDescs(fieldDesc.ParentFile()),
+			cel.TypeDescs(fieldDesc.Message().ParentFile()),
```
oliversun9, line 271: This is subtle indeed 😂

oliversun9 (approved):
LGTM, thanks for fixing it!

PR #56, opened 2023-10-02
Move package `celext` out of internal

oliversun9:
Move `celext` out of internal so that `buf lint` can import this package and load `DefaultEnv` before checking if a CEL expression compiles.


celext/lib.go
rodaine, line 1: Can you remove the reference to "locale" from the DefaultEnv comment? It's a vestigial bit of configuration that was removed prior to release.

oliversun9: Resolved in [147fd32](https://github.com/bufbuild/protovalidate-go/pull/56/commits/147fd3254d49016a748cffc008e5cc9eb94afdb7), also checked that this is the only place referencing "locale"

rodaine (approved):

PR #59, opened 2023-10-03
Make DefaultResolver public

oliversun9:
Make DefaultResolver public so that `buf lint` can use it to resolve `MessageConstraints` and `FieldConstraints`.


internal/evaluator/builder.go
```diff
@@ -38,6 +38,12 @@ type Builder struct {
 	Load        func(desc protoreflect.MessageDescriptor) MessageEvaluator
 }
 
+type StandardConstraintResolver interface {
```
oliversun9, line 41: This was from `internal/evaluator/resolver.go`, but the rest of that file is moved out of `internal`.

rodaine (approved):

PR #60, opened 2023-10-16
Allow passing a custom env for a validator

simon0191:
Quick and dirty PR aiming for some feedback :)

---

Allow passing a custom env for a validator so that the caller can include custom CEL functions.

For example:

```go

env, err := celext.DefaultEnv(true, celext.WithEnvOptions(
   cel.Function("isUUID",
			cel.MemberOverload(
				"string_is_uuid_bool",
				[]*cel.Type{cel.StringType},
				cel.BoolType,
				cel.FunctionBinding(func(value ref.Val) ref.Val {
                                          ....
				}),
			),
		),
))
assert.NoError(err)
validator := protovalidate.New(protovalidate.WithEnv(env))

```

### DIFF
--- a/celext/lib.go
+++ b/celext/lib.go
@@ -30,15 +30,22 @@ import (
 	"github.com/google/cel-go/ext"
 )
 
+var _ cel.Library = lib{}
+
 // DefaultEnv produces a cel.Env with the necessary cel.EnvOption and
 // cel.ProgramOption values preconfigured for usage throughout the
 // module. If useUTC is true, timestamp operations use the UTC timezone instead
-// of the local timezone.
-func DefaultEnv(useUTC bool) (*cel.Env, error) {
+// of the local timezone. If locale is non-empty, the provided locale string is
+// used for string formatting, defaulting to 'en_US' if unset.
+func DefaultEnv(useUTC bool, options ...EnvOption) (*cel.Env, error) {
+	l := lib{
+		useUTC: useUTC,
+	}
+	for _, option := range options {
+		option(&l)
+	}
 	return cel.NewEnv(
-		cel.Lib(lib{
-			useUTC: useUTC,
-		}),
+		cel.Lib(l),
 	)
 }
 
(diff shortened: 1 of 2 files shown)


rodaine:
I'm assuming your example above is just demonstrative, but just in case it's not, there is a [standard constraint for UUIDs](https://github.com/bufbuild/protovalidate/blob/main/proto/protovalidate/buf/validate/validate.proto#L2816) already.

birdayz:
@rodaine we have a use case where we want to run certain validations only for Update calls in gRPC, but not for create calls. in this case, we cannot use standard constraints, because they would apply to all validations, as the field is present in both Update and Create calls.

so we add this as CEL expression to UpdateXYZ request, so it only applies to the Update call.

rodaine:
Ah, that makes sense! We still want to make sure that everything remains self-contained within the proto files + the agreed upon functions required by the protovalidate spec, so we still likely won't accept this change.

If it's an option, another strategy you could use for this is have create/update request messages only include fields that can actually be mutated by the caller and only return the resource's message in responses. In my experience, this makes the semantics of the types clearer (there's no ambiguity about what a user can set vs what's computed by the server).

That said, I can see a feature where instead of just `skipped` at the field level, we can support `skipped_fields` which takes a FieldMask and can specify which fields _not_ to evaluate. I think that'd work in your usecase. Then, you can have the uuid constraint on the field normally, and disable it for circumstances where you don't want to. This means the message describes its constraints accurately (the uuid field must be a uuid), but loosens it as an exception.

simon0191:
>We are, however, https://github.com/bufbuild/protovalidate/issues/51 so that they can be applied easily to multiple fields without repeating the expression logic in N places.

This would be really useful.

>another strategy you could use for this is have create/update request messages only include fields that can actually be mutated by the caller and only return the resource's message in responses. In my experience, this makes the semantics of the types clearer (there's no ambiguity about what a user can set vs what's computed by the server).

Good alternative. We'll give this a try.

>I can see a feature where instead of just skipped at the field level, we can support skipped_fields which takes a FieldMask and can specify which fields not to evaluate. I think that'd work in your usecase. Then, you can have the uuid constraint on the field normally, and disable it for circumstances where you don't want to. This means the message describes its constraints accurately (the uuid field must be a uuid), but loosens it as an exception.

This also sounds like a good addition

---

Thanks for the feedback @rodaine

PR #65, opened 2023-10-20
Fix bug where cel expression cannot compile for fields of type google.protobuf.Any

oliversun9:
This fixes the problem where cel expressions defined on fields of type `google.protobuf.Any` cannot compile.

For example, when validating a `FieldOfTypeAny` with its `any` field set to a `Simple`, it would fail with 
```
runtime error: error evaluating any_type: unknown type: 'tests.example.v1.Simple'
```
```
message Simple {
  string s = 1;
}

message FieldOfTypeAny {
  google.protobuf.Any any = 1 [(buf.validate.field).cel = {
    id: "any_type",
    message: "this should never fail",
    expression: "this == this"
  }];
}
```


internal/evaluator/builder.go
```diff
@@ -274,7 +274,7 @@ func (bldr *Builder) processFieldExpressions(
 	var opts []cel.EnvOption
 	if fieldDesc.Kind() == protoreflect.MessageKind {
 		opts = []cel.EnvOption{
-			cel.Types(dynamicpb.NewMessage(fieldDesc.Message())),
+			cel.Types(dynamicpb.NewMessage(fieldDesc.ContainingMessage())),
```
rodaine, line 277: Let's add both in this case.

oliversun9: resolved in [5a9fae1](https://github.com/bufbuild/protovalidate-go/pull/65/commits/5a9fae13ee69efa176b7471a672e09dfb1624810)

rodaine (approved):

PR #70, opened 2023-10-26
Use make lint in CI instead of golangci-lint action

oliversun9:
Run `make lint` in CI instead of using the action to make sure CI is consistent with running `make lint` locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ lint: lint-proto lint-go  ## Lint code and protos
 
 .PHONY: lint-go
 lint-go: $(BIN)/golangci-lint
-	$(BIN)/golangci-lint run ./...
+	$(BIN)/golangci-lint run --modules-download-mode=readonly --timeout=3m0s ./...
 
 .PHONY: lint-proto
 lint-proto: $(BIN)/buf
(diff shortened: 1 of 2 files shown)


rodaine (approved):

akshayjshah:
Can we just run `make lint` instead of making this extra-complicated (and different from our other OSS repos)?

Makefile
```diff
@@ -38,7 +38,7 @@ lint: lint-proto lint-go  ## Lint code and protos
 
 .PHONY: lint-go
 lint-go: $(BIN)/golangci-lint
-	$(BIN)/golangci-lint run ./...
+	$(BIN)/golangci-lint run --modules-download-mode=readonly --timeout=3m0s ./...
```
akshayjshah, line 41: Huh, this is interesting! @oliversun9 can you also update connectrpc/connect-go and bufbuild/skeleton-oss-go to use this flag? I think those are the repositories we usually use as templates, so if we update them the change will eventually spread to the other repos.

oliversun9: Cool! They are here: https://github.com/connectrpc/connect-go/pull/617 and https://github.com/bufbuild/skeleton-oss-go/pull/8

PR #72, opened 2023-10-26
Run CI on Go 1.19

akshayjshah:


nicksnyder:
Unless you are going to get rid of atomic.Pointer, 1.19 is the minimum version (and we should explicitly add that to the test matrix instead of 1.18.

If you are going to remove atomic.Pointer reference so this project does support 1.18, then the diff should include updating the go.mod file too.

akshayjshah:
> Unless you are going to get rid of atomic.Pointer, 1.19 is the minimum version (and we should explicitly add that to the test matrix instead of 1.18).
> 
> If you are going to remove atomic.Pointer reference so this project does support 1.18, then the diff should include updating the go.mod file too.

Forgot that we're all the way to Go 1.21 these days :/ Getting rid of `atomic.Pointer` is a one-liner, but supporting stable, oldstable, and old-oldstable feels like enough.

Amended to move the minimum CI version to 1.19.

nicksnyder (approved):

PR #75, opened 2023-11-01
Check whether scalar fields are empty in (buf.validate.oneof).required

oliversun9:
We used to check whether a scalar field is _present_ when enforcing `(buf.validate.oneof).required`,
while checking whether a scalar field is not _empty_ when enforcing `(buf.validate.field).required`.

This PR updates `(buf.validate.oneof).required`'s semantics to check a scalar field is not empty (as opposed to present) in `(buf.validate.oneof).required`.

Conformance tests are updated in https://github.com/bufbuild/protovalidate/pull/123


internal/evaluator/oneof.go
```diff
@@ -33,12 +33,26 @@ func (o oneof) Evaluate(val protoreflect.Value, failFast bool) error {
...
+		FieldPath:    string(o.Descriptor.Name()),
+		ConstraintId: "required",
+		Message:      "exactly one field is required in oneof",
+	}}}
+	presentFieldDescriptor := msg.WhichOneof(o.Descriptor)
+	if presentFieldDescriptor == nil {
+		return validationErr
+	}
+	if presentFieldDescriptor.Kind() == protoreflect.GroupKind || presentFieldDescriptor.Kind() == protoreflect.MessageKind {
+		if !msg.Get(presentFieldDescriptor).Message().IsValid() {
```
oliversun9, line 49: `IsValid` checks [this case](https://github.com/bufbuild/protovalidate/pull/123#discussion_r1379033039)

PR #80, opened 2023-11-17
Resolve correct CEL types consistently for all field types

rodaine:
The way we were resolving the CEL types for type-checking expressions was inconsistent between custom and standard constraints. Resulting in compilation errors (particularly around repeated fields). Logic shared between the two (mostly lookups) was moved into the `expression` internal package and used uniformly for both environments.

This also improves on a previously discovered bug around the Any WKT where custom expressions against such a field would fail with a runtime error if its underlying type was not known to CEL (CEL treats Any's as the underlying type, instead of the Any message itself). The standard constraints on Any do not have this limitation. We are populating the root CEL environment with `protoregistry.GlobalFiles` for now, but will likely make this configurable in the long-run.

Context: https://github.com/bufbuild/protovalidate/issues/92#issuecomment-1813198938 (h/t @matthewpi)


internal/constraints/cache.go
```diff
@@ -114,7 +114,7 @@ func (c *Cache) prepareEnvironment(
 ) (*cel.Env, error) {
 	env, err := env.Extend(
 		cel.Types(rules.Interface()),
-		cel.Variable("this", c.getCELType(fieldDesc, forItems)),
+		cel.Variable("this", expression.ProtoFieldToCELType(fieldDesc, true, forItems)),
```
oliversun9, line 117: For my own knowledge, why is generic set to true here?

oliversun9 (approved):
LGTM, thanks for doing this! Is it possible to also export the `expression` package so that the CLI can reuse it for linting CEL expressions?

internal/constraints/cache.go
rodaine: Here we are compiling the standard constraint expressions the first time we see them, and we use a concrete field it's attached to as the basis for type information. `repeated` and `map` fields are generic over their elements, so when we build (say) the `unique` expression for `RepeatedRules`, we cannot assume that the element type of the list is always going to be that of the field we're using since another repeated field of a different type might also specify unique.

rodaine:
> LGTM, thanks for doing this! Is it possible to also export the `expression` package so that the CLI can reuse it for linting CEL expressions?

I'm sure we could!

PR #83, opened 2023-11-18
Move out of internal functions that create env options for field descriptors

oliversun9:
Move `RequiredCELEnvOptions` and `ProtoFieldToCELType` out of internal so that it can be reused by the buf CLI for `buf lint`.

Also rename `ProtoKindToCELType` to `protoKindToCELType` because no other packages calls it directly.

### DIFF
--- a/internal/constraints/lookups_test.go
+++ b/internal/constraints/lookups_test.go
@@ -17,7 +17,7 @@ package constraints
 import (
 	"testing"
 
-	"github.com/bufbuild/protovalidate-go/internal/expression"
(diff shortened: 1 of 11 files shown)


internal/constraints/lookups_test.go
```diff
@@ -54,37 +52,3 @@ func TestExpectedWrapperConstraints(t *testing.T) {
 		})
 	}
 }
-
-func TestProtoKindToCELType(t *testing.T) {
```
oliversun9, line 58: Moved to `celext/lookups_test.go`

rodaine (approved):

PR #84, opened 2023-11-28
Fix email validation

dragon3:
This pull request fixes email validation (`celext.validateEmail`).

I found that some unexpected strings like `foo@example.com    ` (with trailing spaces) got allowed by the email validation, those should not be valid.

`mail.ParseAddress` can parse the string without any errors and the result's address (`a.Address` ) is valid, but it's already different from the original string, so I think checking if `a.Address` and the original `addr` is same is necessary.


rodaine (approved):
Thanks for the patch, @dragon3! This looks great. 😁

dragon3:
Thank you! 🎉

PR #95, opened 2024-02-12
Support string.host_and_port standard constraint

rodaine:


celext/lib.go
```diff
@@ -467,3 +486,36 @@ func (l lib) validateIPPrefix(p string, ver int64, strict bool) bool {
 		return false
 	}
 }
+
+func (l lib) isHostAndPort(val string, portRequired bool) bool {
+	splitIdx := strings.LastIndexByte(val, ':')
+
+	if val[0] == '[' { // ipv6
```
pkwarren, line 493: Should this verify the string isn't empty first?

pkwarren (approved):

celext/lib.go
rodaine: oh yes for sure 😬

pkwarren (approved):

PR #102, opened 2024-02-26
Update license years and fix makefile

pkwarren:
Update license years to include 2024. Fix makefile so we rely on the Go compiler to keep $(BIN)/protovalidate-conformance-go updated instead of make rules (which weren't causing it to be rebuilt when changes to .go files were made). Update generated code to protobuf-go 1.32.0 to match go.mod requirement.


Makefile
```diff
@@ -94,12 +95,12 @@ $(BIN)/license-header: $(BIN) Makefile
...
+		github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION)
 
 $(BIN)/protovalidate-conformance: $(BIN) Makefile
 	GOBIN=$(abspath $(BIN)) $(GO) install \
     	github.com/bufbuild/protovalidate/tools/protovalidate-conformance@$(CONFORMANCE_VERSION)
 
-$(BIN)/protovalidate-conformance-go: $(BIN) Makefile
-	GOBIN=$(abspath $(BIN)) $(GO) build -o $(BIN)/protovalidate-conformance-go \
-		./internal/cmd/protovalidate-conformance-go
+.PHONY: protovalidate-conformance-go
```
pkwarren, line 104: Just made this a PHONY target - the go compiler is super fast and this will ensure we always keep the binary updated when running targets like `make conformance`.

rodaine (approved):

PR #114, opened 2024-04-02
Add tuuid conformance tests

rodaine:


go.sum
```diff
@@ -1,5 +1,7 @@
 buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.33.0-20240221180331-f05a6f4403ce.1 h1:0nWhrRcnkgw1kwJ7xibIO8bqfOA7pBzBjGCDBxIHch8=
 buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.33.0-20240221180331-f05a6f4403ce.1/go.mod h1:Tgn5bgL220vkFOI0KPStlcClPeOJzAv4uT+V8JXGUnw=
+buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.33.0-20240401165935-b983156c5e99.1 h1:2IGhRovxlsOIQgx2ekZWo4wTPAYpck41+18ICxs37is=
```
pkwarren, line 3: It looks like `go mod tidy` cleans up a lot of this file - perhaps we should do that as part of `generate` so `checkgenerate` will fail the build if there are changes?

rodaine: Wouldn't hurt. Follow up PR though.

pkwarren (commented):
When I build locally with `make` i get an updated `strings.pb.go` created but that isn't showing up in this PR - is that expected?

rodaine:
Ah should have run generate. The conformance executor doesn't use the generated protos at all so it's not a big deal (I occassionally use them for regression tests in the codebase though), but we probably need to add a generate check test to this repo for sure.

pkwarren (approved):

PR #118, opened 2024-04-16
feature: add option to extend CEL environment

igor-tsiglyar:
Adding an option to extend the CEL environment used for validation with custom functions or constants. This would greatly improve the protovalidate usage in dynamic environment (e.g. when some of the constraints are not static and hence cannot be defined using existing rules like gt, min_len etc.). This would also allow for custom format support, e.g. the formats that are not widely used and thus not making sence to be supported at celext level.


rodaine:
Hi, @igor-tsiglyar! Thanks for the patch. However, we are not going to expose the CEL environment for extension. For a previous discussion on this topic (and others), see https://github.com/bufbuild/protovalidate-go/issues/73#issuecomment-1792999337

PR #119, opened 2024-04-20
add ability to work with custom cel env

ozanturksever:


rodaine:
Hi, @ozanturksever! Thanks for the patch. However, we are not going to expose the CEL environment for extension. For a previous discussion on this topic (and others), see https://github.com/bufbuild/protovalidate-go/issues/73#issuecomment-1792999337

PR #124, opened 2024-05-20
Update buf CLI to 1.32 + migrate

rodaine:


buf.gen.yaml
```diff
@@ -1,12 +1,15 @@
...
+    - file_option: go_package
+      module: buf.build/envoyproxy/protoc-gen-validate
+    - file_option: go_package
+      module: buf.build/bufbuild/protovalidate
+  override:
+    - file_option: go_package_prefix
+      value: github.com/bufbuild/protovalidate-go/internal/gen
 plugins:
-  - plugin: buf.build/protocolbuffers/go:v1.32.0
+  - remote: buf.build/protocolbuffers/go:v1.32.0
```
pkwarren, line 13: Do you want to bump this to a newer version?

buf.yaml
```diff
@@ -0,0 +1,23 @@
...
+lint:
+  use:
+    - DEFAULT
+  except:
+    - FIELD_NOT_REQUIRED
+    - PACKAGE_NO_IMPORT_CYCLE
+  ignore_only:
+    PROTOVALIDATE:
+      - proto/tests/example/v1/validations.proto
+  disallow_comment_ignores: true
```
pkwarren, line 17: This is part of `buf config migrate` for compatibility sake, but I'd remove this setting and the `except:` setting to migrate to the defaults (as if you had created a new v2 config).

buf.yaml
```diff
@@ -0,0 +1,23 @@
...
+breaking:
+  use:
+    - FILE
+  except:
```
pkwarren, line 21: I'd also remove these to pick up the new v2 defaults.

buf.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - buf.build/envoyproxy/protoc-gen-validate
+lint:
+  use:
+    - DEFAULT
+  except:
+    - FIELD_NOT_REQUIRED
+    - PACKAGE_NO_IMPORT_CYCLE
+  ignore_only:
+    PROTOVALIDATE:
+      - proto/tests/example/v1/validations.proto
```
pkwarren, line 16: Is this meant to be relative to the module path or relative to the buf.yaml file?

buf.gen.yaml
rodaine: Suppose I could since I'm in here

pkwarren (approved):

PR #128, opened 2024-06-25
Small improvement to unique implementation

pkwarren:
Update the unique implementations to allocate a sufficiently sized map given the input to avoid reallocations. Avoid allocations if the size is less than or equal to 1.

### DIFF
--- a/celext/lib.go
+++ b/celext/lib.go
@@ -378,12 +378,20 @@ func (l lib) uniqueMemberOverload(itemType *cel.Type, overload func(lister trait
 }
 
 func (l lib) uniqueScalar(list traits.Lister) ref.Val {
-	exist := make(map[ref.Val]struct{})
-	for i := int64(0); i < list.Size().Value().(int64); i++ {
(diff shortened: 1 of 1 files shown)


celext/lib.go
```diff
@@ -378,12 +378,20 @@ func (l lib) uniqueMemberOverload(itemType *cel.Type, overload func(lister trait
 }
 
 func (l lib) uniqueScalar(list traits.Lister) ref.Val {
-	exist := make(map[ref.Val]struct{})
-	for i := int64(0); i < list.Size().Value().(int64); i++ {
-		if _, ok := exist[list.Get(types.Int(i))]; ok {
+	size, ok := list.Size().Value().(int64)
+	if !ok {
+		return types.UnsupportedRefValConversionErr(list.Size().Value())
```
pkwarren, line 383: Don't think this is really reachable but golangci-lint complained so I added error handling.

rodaine (approved):

PR #138, opened 2024-08-16
Add `buf.gen.yaml` v2 schema example

milas:
This took me a bit to figure out the right invocation, so I figured it'd be useful to others as well.


rodaine (approved):
Thank you, @milas!

PR #139, opened 2024-08-16
Upgrade to support Go 1.23

rodaine:


.github/workflows/ci.yaml
```diff
@@ -18,7 +18,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [ 1.20.x, 1.21.x, 1.22.x ]
+        go-version:
+          - 1.20.x
```
jhump, line 22: Remove 1.20? I think that's what we've done in other repos. We generally support back to N-2 (one farther back than the Go team).

.github/workflows/ci.yaml
```diff
@@ -33,8 +37,5 @@ jobs:
       - name: Lint
         # Often, lint guidelines depend on the Go version. To prevent
         # conflicting guidance, run only on the most recent supported version.
-        if: matrix.go-version == '1.22.x'
-        run: make lint
-      - name: Check Generate
-        if: matrix.go-version == '1.22.x'
-        run: make checkgenerate
```
jhump, line 40: Why remove this check?

rodaine: It exists in the buf tests already, but I do need to make sure the buf tests run for every PR (instead of only on proto changes).

.github/workflows/ci.yaml
rodaine: We don't really need to ratchet it forward to 1.21, but I can.

rodaine: I need to disable the 1.20 required check first, so will merge this as-is and follow up with a PR to remove it.

.github/workflows/ci.yaml
rodaine: Done in https://github.com/bufbuild/protovalidate-go/pull/139/commits/bd22c50bb5e446a9c50286701444ee21692543ff

rodaine:
Once this merges, I'll make the buf & v1.23 checks required, make v1.20 no longer required, and a follow up PR to remove v1.20 from the matrix.

jhump (approved):

PR #142, opened 2024-08-29
Bugfix: avoid spurious tautology pruning when dealing with certain recursive messages

### LINKED ISSUE #141: [BUG] Recursive validation skipped depending on field order
## Description

A bit difficult to describe, but there are circumstances where the validator will incorrectly skip a message.

https://github.com/daemonl/pvbug is a full demo

Validating an empty 'FieldWithIssue' gives a validation error, correctly.

When nested as `OneTwo.F1.field`, it also gives the validation error.

However, when nested as `TwoOne.F1.field`, and when the `need_this` string is set to a value, it does not fail validation.

The number of the fields doesn't seem to matter, only the order they appear in.

```proto
message OneTwo {
  F1 field1 = 1;
  F2 field2 = 2;
}

message TwoOne {
  F2 field2 = 1;
  F1 field1 = 2;
}

message F2 {
  FieldWithIssue field = 1;
}

message F1 {
  string need_this = 1; // This needs a value, leaving it empty validates correctly
  FieldWithIssue field = 2;
}

message FieldWithIssue {
  F1 f1 = 1; // This must be before the name string, but does not need a value
  string name = 2 [(buf.validate.field).string.min_len = 1];
}
```

## Steps to Reproduce

```bash
$ git clone https://github.com/daemonl/pvbug
$ cd pvbug
$ buf generate && go test ./...
```

## Expected Behavior

Expected to see the message marked as invalid in all cases

## Actual Behavior

```
    --- FAIL: TestLocalJ5 (0.01s)
    --- FAIL: TestLocalJ5/Wrapped_in_order_2_1 (0.00s)
        test_test.go:75: Fresh Validation should have failed
FAIL
FAIL    github.com/daemonl/pvbug        0.271s
FAIL
```

## Environment

- **Operating System**:  macOS - M1
- **Version**: macOS Sonoma 14.6.1
- **Compiler/Toolchain**: go1.22.4 darwin/arm64
- **Protobuf Compiler & Version**: buf 1.33.0
- **Protoc-gen-validate Version**: ?
- **Protovalidate Version** v0.6.4

## Possible Solution

I'm gonna guess it's something to do with recursive lazy loading

rodaine:
When protovalidate-go builds up evaluators for a message, it first adds the pointer to the evaluator cache so any fields that recursively include that message will use that instead of attempting to rebuild the message infinitely until a stack overflow. 

Also as part of this build process, any sub-evaluators that would tautologically validate (i.e., always be valid) are pruned to optimize the resulting evaluator for the message. Generally speaking, if an evaluator has no sub-evaluators or CEL expressions attached to it, then it's considered a tautology.

Individually, these features work well, but together they could collide causing validations to be lost for certain recursive message structures. For example, with the following message graph (arrows meaning "has a field of type"):

```mermaid
  graph TD;
      M1-->M2;
      M2-->M3;
      M3-->M2;
      M1-->M3;
```

1. Suppose in this case, the only constraint is defined on `M2`. 
2. `M1` starts to build, and as part of that, it starts to build `M2`. 
3. This continues to `M3`.
4. `M3` recursively depends on `M2`, and ends up loading the still incomplete pointer to `M2`'s evaluator from the cache. 
5. Before adding it to `M3`'s list of sub-evaluators, we check if it's a tautology.
6. Depending on field ordering, `M2`'s evaluator might have no sub-evaluators yet for its constraint, resulting in a false positive.
7. This means `M3` would not include the `M2` sub-evaluator.
8. Bubbling up back to `M2`, it would also check if `M3`'s evaluator is a tautology, and since `M2`'s was excluded, `M3` is also a tautology false positive, and won't be included in `M2`'s sub-evaluators. 
9. Then, however, `M2` will pick up its actual constraints, resulting in `M1` not detecting a tautology. 
10. Finally, `M1` pulls `M3`'s evaluator from the cache, sees that it is (wrongly) a tautology, and does not include it in its sub-evaluators.

---

Originally, I attempted to make the tautology detection check if the message's evaluator was "ready," but this resulted in infinite recursion when checking tautology between the messages. For now, instead, we treat all message evaluators as never a tautology. This solves the underlying issue, at the cost of nominally less optimal evaluators. This is certainly acceptable in that any other solution will likely make the evaluator builder more complex than it already is.

---

Longer term solutions to explore:
- Track recursion and short-circuit when detected. Similar to this solution, but would only occur in recursive scenarios.
- Do tautological pruning as a second pass after the evaluator has been completely constructed. Less efficient than doing it as we go, but will ensure that everything is built prior to the check.
- Leave it as-is. The impact on memory and CPU is minimal.

Fixes #141

### DIFF
--- a/internal/evaluator/builder.go
+++ b/internal/evaluator/builder.go
@@ -83,7 +83,7 @@ func (bldr *Builder) load(desc protoreflect.MessageDescriptor) MessageEvaluator 
 }
 
 // loadOrBuild either returns a memoized MessageEvaluator for the given
-// descriptor, or lazily constructs a new one. This method is thread-safe via
+// descriptor or lazily constructs a new one. This method is thread-safe via
 // locking.
 func (bldr *Builder) loadOrBuild(desc protoreflect.MessageDescriptor) MessageEvaluator {
 	if eval, ok := (*bldr.cache.Load())[desc]; ok {
--- a/internal/evaluator/message.go
+++ b/internal/evaluator/message.go
@@ -41,7 +41,12 @@ func (m *message) EvaluateMessage(msg protoreflect.Message, failFast bool) error
 }
 
 func (m *message) Tautology() bool {
-	return m.Err == nil && m.evaluators.Tautology()
+	// returning false for now to avoid recursive messages causing false positives
+	// on tautology detection.
+	//
+	// TODO: use a more sophisticated method to detect recursions so we can
+	//  continue to detect tautologies on message evaluators.
+	return false
 }
 
 func (m *message) Append(eval MessageEvaluator) {
--- a/internal/gen/tests/example/v1/validations.pb.go
+++ b/internal/gen/tests/example/v1/validations.pb.go
@@ -746,6 +746,273 @@ func (x *RepeatedItemCel) GetPaths() []string {
 	return nil
 }
 
+type OneTwo struct {
+	state         protoimpl.MessageState
+	sizeCache     protoimpl.SizeCache
+	unknownFields protoimpl.UnknownFields
+
+	Field1 *F1 `protobuf:"bytes,1,opt,name=field1,proto3" json:"field1,omitempty"`
+	Field2 *F2 `protobuf:"bytes,2,opt,name=field2,proto3" json:"field2,omitempty"`
+}
+
+func (x *OneTwo) Reset() {
+	*x = OneTwo{}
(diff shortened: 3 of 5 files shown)


daemonl:
Can confirm this branch fixes both my initial case and the example case! Thanks heaps!

jchadwick-buf (approved):

PR #144, opened 2024-09-04
Implement predefined field constraints

jchadwick-buf:
This fails in CI for now, it depends on https://github.com/bufbuild/protovalidate/pull/246.

Assuming you have `protovalidate` cloned in the parent folder of `protovalidate-go`, and that you've ran `make conformance` after checking out https://github.com/bufbuild/protovalidate/pull/246, the following command can be used to test against the local protovalidate. This will not currently pass all the way; it will have two failures that are fixed by https://github.com/bufbuild/protovalidate/pull/245.

```sh
(rm -fr .tmp/protovalidate; buf generate ../protovalidate/proto/protovalidate -o .tmp/protovalidate; mv .tmp/protovalidate/internal/gen/buf .tmp/protovalidate; rm -fr .tmp/protovalidate/internal; cd .tmp/protovalidate; go mod init buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go; go mod tidy; cd ..; go mod edit -replace=buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go=./.tmp/protovalidate && go mod tidy) && go build -o protovalidate-conformance-go ./internal/cmd/protovalidate-conformance-go && ../protovalidate/.tmp/bin/protovalidate-conformance --strict --strict_message --strict_error ./protovalidate-conformance-go
```


internal/constraints/cache.go
```diff
@@ -57,8 +61,19 @@ func (c *Cache) Build(
...
-		precomputedASTs, compileErr := c.loadOrCompileStandardConstraint(env, desc)
+	constraints.Range(func(desc protoreflect.FieldDescriptor, rule protoreflect.Value) bool {
+		fieldEnv, compileErr := env.Extend(
+			cel.Constant(
+				"rule",
+				celext.ProtoFieldToCELType(desc, true, forItems),
+				types.DefaultTypeAdapter.NativeToValue(rule.Interface()),
+			),
+		)
+		if err != nil {
```
oliversun9, line 72: Should this be `if compileErr != nil {` instead?

jchadwick-buf: Good call, confirmed and fixed.

jchadwick-buf:
FYI: Some changes are being worked on with regards to compatibility. For posterity sake I want to write an explanation. In particular, I noticed some rather serious problems with how our approach to resolving unknown extensions would interplay with users that need to use dynamic descriptor sets; we were relying on the GlobalTypes registry to resolve extensions, but this would not work for users that many have multiple sets of file descriptor sets that are distinct, as the user will not be able to register them. To rectify this I unified the code that resolves extensions for all constraint options.

Now there are tests that should ensure _any_ `proto.Message` with field constraints will always resolve, no matter if they're using the concrete gencode types, somehow an incompatible gencode type (e.g. if the protovalidate package path changes), a `dynamicpb.Message`, or if the field constraints are hidden away in unknown fields. In addition, to ensure predefined field constraints are resolved correctly, the user can pass in an extension resolver when creating a validator. Otherwise, `GlobalTypes` will be used, which works in all of the obvious cases. Similar adaptations will probably be needed for other versions (I already added it to the Java runtime.)

There's actually quite a few bits of nuance I ran into that made this tricky. For example, we cannot mutate the return value of `Options()` - it is illegal to do so according to protoreflect and would not be goroutine safe anyway, which would probably surprise people since it would only ever come up in edge cases.

Despite this, we do have at least one win: it is possible to avoid `RangeExtensions` while still not `panic`ing, because we can use `ProtoReflect().Get()` instead, which is exactly what `proto.GetExtension` does internally anyways. Then later on, we can use the same logic we normally would to determine if a reparse is necessary.

The code winds up having to do up to four field presence checks:
- On the concrete fields, using the normal extension tag number
- On the concrete fields, using the legacy extension tag number
- On the reparsed unknown fields, using the normal extension tag number
- On the reparsed unknown fields, using the legacy extension tag number

The reparsed checks have to be done separately because again, we can't actually mutate the field options, so we have to reparse into a new message and check that. However this should not be a major concern, since this step is cached.

A net effect of this is that the conformance test suite runner is even more agnostic to the input data, and previous versions of the conformance test suite should work correctly with the new runner.

That said, at the moment this is only _backward_ compatibility and not forward compatibility. If you use a standard constraint that is too new, it will not work, just as it currently does not work. I am going forward with the assumption that this is OK but it may warrant some thought so I'll ask a couple folks their thoughts anyways.

jchadwick-buf:
Update: Added a `WithAllowUnknownFields` option which defaults to `false`. That way, the default behavior doesn't allow missing definitions to result in silently-broken validation. This is another breaking change, although in theory there is no way it can break anyone, because the version that adds it will also be the version that adds custom predefined rules, and by definition has a superset of all standard constraints ever defined prior.

I have also been investigating to ensure that our changes do not tank performance. Performance is mostly within margin of error although the switch to proto2 has increased the number of allocations in the invalid path (because the violation structure with proto2 explicit field presence is full of pointer-chasing.) I think it isn't worth investing any time into. Not sure what you could really do to improve that; string interning would probably help but I think it would be extremely unlikely to matter in real world use cases.

jchadwick-buf:
Unfortunately I am currently blocked on getting the upstream protovalidate changes released for the moment so I have made some temporary adaptations in 040b31c2d53e1bafbd79c8f362c9a419cf72253f that will make the diff as close to possible as the actual final diff. I'm hoping that we can get the protovalidate changes released before end-of-week but getting some eyes on the changes here would be great either way.

jchadwick-buf:
Protovalidate 0.8.1 is out now so I've reverted the temporary changes and updated to it. We should be good to go.

internal/constraints/cache.go
```diff
@@ -170,3 +199,20 @@ func (c *Cache) getExpectedConstraintDescriptor(
 		return expected, ok
 	}
 }
+
+func reparseUnrecognized(
```
oliversun9, line 203: We [added a resolver to this function](https://github.com/bufbuild/buf/pull/3317/files#diff-85d4d81d3a6dfea74ad1d6a24a3bbc1654253cd5632e8e0fe3ae7618433bf2d0R78-R82) with synchronicity and same parameter name as well 😄

oliversun9: but yea i'm guessing I don't need the copy of the function any more since I should be to a resolver to create the validator. Either way, it's not a concern of this PR.

internal/constraints/cache.go
```diff
@@ -44,21 +46,45 @@ func (c *Cache) Build(
...
+	)
 	if done {
 		return nil, err
 	}
 
+	if err = reparseUnrecognized(extensionTypeResolver, constraints); err != nil {
+		return nil, errors.NewCompilationErrorf("error reparsing message: %w", err)
+	}
+	if !allowUnknownFields && len(constraints.GetUnknown()) > 0 {
+		return nil, errors.NewCompilationErrorf("unknown constraints in %s", constraints.Descriptor().FullName())
```
oliversun9, line 66: Do we also want to suggest the `WithAllowUnknownFields` option in the error message?

validator.go
```diff
@@ -183,3 +191,25 @@ func WithStandardConstraintInterceptor(interceptor StandardConstraintInterceptor
...
+	}
+}
+
+// WithAllowUnknownFields specifies if the presence of unknown field constraints
+// should cause compilation to fail with an error. When set to false, an unknown
+// field will simply be ignored, which will cause constraints to silently not be
+// applied. This condition may occur if a predefined constraint definition isn't
+// present in the extension type resolver, or when passing dynamic messages with
+// standard constraints defined in a newer version of protovalidate. The default
+// value is false, to prevent silently-incorrect validation from occurring.
```
oliversun9, line 210: I'm a little worried by the `false` default value. I think this means every time a new user-defined rule is added as an extension and added to some field, the side of transportation that validates the payload would need to update the generated Go code, before the other side does. But I could have confused myself there and perhaps this is only a problem with dynamic messages (also not sure what I said made much sense)

jchadwick-buf: This should only impact dynamic file descriptor sets. Field constraints are specified in field options, and field options are stored in the file descriptor set, not on the protobufs themselves, so it only has to be consistent with itself. This is basically only a problem with dynamically-loaded file descriptor sets.

The reason for defaulting this to `false`:
- It won't break any _existing_ users, since the release that contains this will have a superset of all standard field constraints so far.
- If we default it to `true`, we can't change it to `false` without potentially breaking users later, which I think means it's now or never.
- Crucially, defaulting to `false` means that the behavior will just be incorrect. It will indeed pass validation, without running the custom predefined constraints at all.

jchadwick-buf:
> > For example, we cannot mutate the return value of Options() - it is illegal to do so according to protoreflect and would not be goroutine safe anyway, which would probably surprise people since it would only ever come up in edge cases.
> 
> It doesn't sound like this PR mutates any return value of `Options()` but I am curious for my own knowledge why protovalidate go needs to mutates this value ever 👀

The reason why it _might_ need to mutate the `Options()` data is because we need to reparse the options. The normal reparse logic we'd use in Go just does a merge-parse directly into the `protoreflect.Message`, but doing so here would be illegal, so we do it into an empty `FieldOptions` instead and then loop over those fields separately.

internal/constraints/cache.go
jchadwick-buf: I added a note in d021bd5a7eb6b3ce628651f891cfbccc7745bc86. To keep it succinct I only mention `WithExtensionTypeResolver` _but_ I add a note to the `WithExtensionTypeResolver` option that points to `WithAllowUnknownFields`.

oliversun9: Nice, thanks

validator.go
oliversun9: > field options are stored in the file descriptor set, not on the protobufs themselves

Yea that's where I confused myself, that was a rookie mistake on my part.

I agree with the reasoning.

internal/extensions/resolve.go
```diff
@@ -0,0 +1,167 @@
...
+
+//nolint:gochecknoglobals // static data, only want single instance
+var resolver = newExtensionResolver()
+
+// Resolve resolves extensions without using [proto.GetExtension], in case the
+// underlying type of the extension is not the concrete type expected by the
+// library. In some cases, particularly when using a dynamic descriptor set, the
+// underlying extension value's type will be a dynamicpb.Message. In some cases,
+// the extension may not be resolved at all. This function handles reparsing the
+// fields as needed to get it into the right concrete message as needed. Resolve
```
rodaine, line 40:
```suggestion
// fields as needed to get it into the right concrete message. Resolve
```

internal/extensions/resolve.go
```diff
@@ -0,0 +1,167 @@
...
+		panic(err)
+	}
+	legacyExtension := dynamicpb.NewExtensionType(fileDescriptor.Extensions().Get(0))
+	resolver.register(legacyExtension)
+	resolver.legacyExtensionMap[extension] = legacyExtension
+}
+
+// resolve handles the majority of extension resolution logic. This will return
+// a proto.Message for the given extension if the message has the tag number of
+// the provided extension (or an equivalent legacy extension.) If there was no
```
rodaine, line 127:
```suggestion
// the provided extension (or an equivalent legacy extension). If there was no
```

rodaine (approved):

internal/extensions/resolve.go
```diff
@@ -0,0 +1,167 @@
...
+	resolver.registerLegacy(validate.E_Oneof)
+	return resolver
+}
+
+// register registers an extension into the resolver's registry. This is only
+// called at init and will panic if it fails.
+func (resolver extensionResolver) register(extension protoreflect.ExtensionType) {
+	if err := resolver.types.RegisterExtension(extension); err != nil {
+		//nolint:forbidigo // this needs to be a fatal at init
+		panic(err)
```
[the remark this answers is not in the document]
oliversun9: Sorry I didn't realize this comment wasn't submitted.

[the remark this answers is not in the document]
jchadwick-buf: I always avoid `panic` where possible, but protovalidate will just be non-functional if these registrations fail. We want this ti fail at init just like broken gencode would (because if this happens we likely have a gencode issue.)

internal/extensions/resolve.go
jchadwick-buf: Fixed in 9d5c4696e205f7a84d0d8ba131dd42d1ea2193fe

internal/extensions/resolve.go
jchadwick-buf: Fixed in 9d5c4696e205f7a84d0d8ba131dd42d1ea2193fe

oliversun9 (approved):

PR #151, opened 2024-10-02
Add ResolvePredefinedConstraints method to DefaultResolver

### LINKED ISSUE #147: [Feature Request] Go API for fetching predefined constraints
**Feature description:** 
protovalidate-go has an API for easily fetching field/message/oneof constraints in the [resolver package](https://pkg.go.dev/github.com/bufbuild/protovalidate-go@v0.7.0/resolver). I propose that we add a similar API for getting [predefined constraints](https://github.com/bufbuild/protovalidate/blob/main/docs/predefined-constraints.md) in a similar way.

I see that there is [code in the cache](https://github.com/bufbuild/protovalidate-go/blob/ca39053c85df0f1d01c9c4bcc1914072e17a6583/internal/constraints/cache.go#L166-L169) to fetch them but it seems like there’s not an exposed way to do something similar.

**Problem it solves or use case:** 
It's useful for me to get protovalidate constraints in order to annotate generated OpenAPI specifications from protobuf. I suppose it may be useful for others as well.

**Proposed implementation or solution:** 
ResolveFieldConstraints, ResolveMessageConstraints and ResolveOneofConstraints could be augmented to resolve the predefined constraints from extensions containing `validate.E_Predefined`.

**Contribution:**
protobuf extensions are pretty crazy to me, but I can help if there's no movement on this after a while.

**Examples or references:** 
https://pkg.go.dev/github.com/bufbuild/protovalidate-go@v0.7.0/resolver

**Additional context:**

jchadwick-buf:
This method is not really needed by protovalidate itself, but it is really convenient for us to implement and really inconvenient for consumers to implement, especially if they want to mimic the reparsing logic of protovalidate-go.

Fixes #147.

### DIFF
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -40,3 +40,11 @@ func (r DefaultResolver) ResolveOneofConstraints(desc protoreflect.OneofDescript
 func (r DefaultResolver) ResolveFieldConstraints(desc protoreflect.FieldDescriptor) *validate.FieldConstraints {
 	return extensions.Resolve[*validate.FieldConstraints](desc.Options(), validate.E_Field)
 }
+
+// ResolvePredefinedConstraints returns the PredefinedConstraints option set for
+// the FieldDescriptor. Note that this value is only meaningful if it is set on
+// a field or extension of a field rule message. This method is provided for
+// convenience.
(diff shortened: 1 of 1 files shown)


oliversun9 (approved):
I think the buf CLI also benefit from this, thanks. Would still defer to @rodaine for the final call.

rodaine (approved):

PR #183, opened 2025-01-31
Validate URI query string encoding

smaye81:
This adds an additional check for validating that the query string of a URI is correctly encoded.

### DIFF
--- a/cel/library.go
+++ b/cel/library.go
@@ -488,11 +488,8 @@ func (l library) validateURI(val string, checkAbs bool) bool {
 			return false
 		}
 	}
-	if _, err := url.ParseQuery(uri.RawQuery); err != nil {
-		return false
-	}
-
-	return true
+	_, err := url.ParseQuery(uri.RawQuery)
+	return err == nil
 }
 
 func (l library) isHostAndPort(val string, portRequired bool) bool {
--- a/cel/library.go
+++ b/cel/library.go
@@ -232,8 +232,7 @@ func (l library) CompileOptions() []cel.EnvOption {
 					if !ok {
 						return types.Bool(false)
 					}
-					uri, err := url.Parse(s)
-					return types.Bool(err == nil && uri.IsAbs())
+					return types.Bool(l.validateUri(s, true))
 				}),
 			),
 		),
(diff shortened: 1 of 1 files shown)


cel/library.go
```diff
@@ -479,6 +477,24 @@ func (l library) validateIPPrefix(p string, ver int64, strict bool) bool {
...
+	if err != nil {
+		return false
+	}
+	if checkAbs {
+		ok := uri.IsAbs()
+		if !ok {
+			return false
+		}
+	}
+	if _, err := url.ParseQuery(uri.RawQuery); err != nil {
```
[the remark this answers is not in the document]
smaye81: Yep, done. Let me know if it needs more detail.

rodaine (approved):

cel/library.go
```diff
@@ -479,6 +477,20 @@ func (l library) validateIPPrefix(p string, ver int64, strict bool) bool {
...
+	uri, err := url.Parse(val)
+	if err != nil {
+		return false
+	}
+	if checkAbs && !uri.IsAbs() {
+		return false
+	}
+
+	// Parse the query string to validate it is formed and encoded properly
+	_, err = url.ParseQuery(uri.RawQuery)
```
jchadwick-buf, line 490: I understand the intent here, but this does an awful lot of work just to check the validity of the encoding. It parses out the query components and constructs a map with them, which we don't need.

It's a bit less succinct, but we can get the exact same validation while doing a lot less work:
```go
	// (Not sure if we want to disallow unescaped semicolons, but ParseQuery does.)
	if strings.Contains(uri.RawQuery, ";") {
		return false
	}
	_, err = url.QueryUnescape(uri.RawQuery)
	return err == nil
```

It'd be even better if we could just validate without parsing and decoding everything, but I argue we at least should avoid building the map if we can.

jchadwick-buf (approved):
I think this is a worthwhile improvement in strictness overall, though I definitely am thinking more strongly by the day that we need to formally define the grammar allowed by each validation rule. Relying on Go's sensibilities is nice for protovalidate-go, but even not considering other languages it has some caveats, including this issue of parsing things just to validate them and throw away the parsed result. But, that's definitely out-of-scope for this PR.

cel/library.go
smaye81: Actually in talking to @timostamm today, we may want to discuss this deeper once his validation is done for protovalidate-ts. We can then make sure all the implementations are behaving the same way since he's following the RFC. We could then add additional conformance checks and formally define the grammar for what we validate.

PR #185, opened 2025-02-07
Implement Filter interface

jchadwick-buf:
Adds `ValidationOption` for options passed to the `Validate` function rather than on creation of a validator.

Adds `WithFilter` option for setting a filter to control which fields are validated.

Adds `Filter` interface:

```go
type Filter interface {
	ShouldValidate(message protoreflect.Message, descriptor protoreflect.Descriptor) bool
}
```

- `ShouldValidate` is called once for a message, once for each of its oneofs, and once for each of its fields. The `message` value is the message itself, and the `descriptor` value is the descriptor of the element being operated on.
- `ShouldValidate` will be called for every oneof and field even if it returns false for any message, field or oneof.
- Sub-messages nested under fields will _not_ be checked if `ShouldValidate` returns false.

This is to satisfy the following usecases
- For `buf lint`, we want to be able to single out an individual field to check examples. This can be done by using `descriptor == fieldDescriptor` as a filter. This can also filter oneof and message constraints, which is desirable.
- For field masking, it is possible for a stateful filter to perform field masking using this interface. The only important behavioral detail for that is that we nest depth-first into submessages. This would allow a user to implement out-of-band field masks.


value.go
```diff
@@ -26,6 +26,9 @@ type value struct {
 	Descriptor protoreflect.FieldDescriptor
 	// Constraints are the individual evaluators applied to a value
 	Constraints evaluators
+	// NestedConstraints are constraints applied to constraints nested under a
```
rodaine, line 29:
```suggestion
	// NestedConstraints are constraints applied to messages nested under a
```

Not sure what you're trying to say here exactly.

validator.go
```diff
@@ -179,3 +189,20 @@ func WithAllowUnknownFields() ValidatorOption {
 		c.allowUnknownFields = true
 	}
 }
+
+type validationConfig struct {
+	failFast bool
```
rodaine, line 194: Do we want to permit setting failFast at Validate call-time? I think that would be a nice flexibility now that it's included in the config.

validator.go
```diff
@@ -179,3 +189,20 @@ func WithAllowUnknownFields() ValidatorOption {
...
+
+// A ValidationOption specifies per-validation configuration. See the individual
+// options for their defaults and effects.
+type ValidationOption func(*validationConfig)
+
+// WithFilter specifies a filter to use for this validation. A filter can
+// control which fields are evaluated by the validator.
+func WithFilter(filter Filter) ValidationOption {
+	return func(c *validationConfig) {
+		c.filter = filter
```
rodaine, line 206:
```suggestion
		c.filter = cmp.Or(filter, nopFilter{})
```

rodaine (commented):
Looking good! A few minor things, plus can you add to the benchmarks and examples?

validator.go
jchadwick-buf: In order to do that we do need a bit of refactoring, since there's no obvious way to implement that as-is. I've chosen to follow what connect-go does, turning both types of options into interfaces, having a union interface for options that can be applied in both circumstances, and then have all of the options implemented as structs. Since this is really cluttering the `validator.go` file I also moved it to `option.go`, also like connect-go.

validator.go
jchadwick-buf: Can't use `cmp.Or` yet, since our module version is go 1.21.1 and this was added in go 1.22. Now that go 1.24 is out though maybe it's time to bump that number. Either way, for now I just did the simpler thing and made a conditional for this.

value.go
jchadwick-buf: The only reason why `NestedConstraints` exists is to make it possible to apply the field-level constraints for a field without applying all of the constraints in nested submessages. Anyway, updated the comment.

rodaine (approved):

PR #199, opened 2025-03-06
feat: FromProto func to get ValidationError from validate.Violations

dro-sh:
Function to map `validate.Violations` to `protovalidate.ValidationError`. I want to use this golang struct instead generated from proto


jchadwick-buf:
Sorry for the delay in a response here. I think we probably don't want to accept this as-is, as it would leave most fields of the `ValidationError` unpopulated; we currently don't have a way to work backwards to populate those fields right now. If this limitation is alright for your use case, then I recommend vendoring a function like this into your own code, but I think we probably don't want to include this upstream as-is. I'll engage the rest of the team to make sure that my judgment is correct here, but I am closing this for now. Sorry :(

PR #201, opened 2025-03-11
Enhance isEmail validation

smaye81:
This enhances the current `isEmail` validation by using a consistent regex that will be used across protovalidate implementations.


cel/library.go
```diff
@@ -420,19 +420,15 @@ func (l library) uniqueBytes(list traits.Lister) ref.Val {
...
-	}
-
-	addr = a.Address
-	if len(addr) > 254 {
-		return false
-	}
-
-	parts := strings.SplitN(addr, "@", 2)
-	return len(parts[0]) <= 64 && l.validateHostname(parts[1])
+	r := regexp.MustCompile("^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$")
```
pkwarren, line 430: We should define this in `NewLibrary` or put it behind a sync.Once/sync.OnceFunc to avoid compiling it each time it is used.

smaye81:
Going to close this and reopen a new one that branches from #202

PR #203, opened 2025-03-11
Enhance `isEmail` validation

smaye81:
This enhances the current isEmail validation by using a consistent regex that will be used across protovalidate implementations.

### DIFF
--- a/cel/library.go
+++ b/cel/library.go
@@ -18,9 +18,9 @@ import (
 	"bytes"
 	"math"
 	"net"
-	"net/mail"
 	"net/netip"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


cel/library_test.go
```diff
@@ -26,7 +27,9 @@ import (
 func TestCELLib(t *testing.T) {
 	t.Parallel()
 
-	env, err := cel.NewEnv(cel.Lib(library{}))
+	env, err := cel.NewEnv(cel.Lib(library{
```
pkwarren, line 30: Should we switch this to use `NewLibrary` as well?

smaye81: I thought that too. Wasn't sure if it was instantiated like this for a reason.

smaye81: Updated. 👍

pkwarren (commented):
Changes look good to me.

cel/library.go
```diff
@@ -35,6 +35,11 @@ import (
 	"google.golang.org/protobuf/types/dynamicpb"
 )
 
+const (
+	// The regular expression to use when validating emails.
```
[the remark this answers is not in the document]
smaye81: Added.

rodaine (approved):

PR #205, opened 2025-03-13
Handle compilation error

### LINKED ISSUE #204: Allow evaluating constraints with `Filter` set even when a `CompilationError` is found
The current behaviour of `CompilationError` in the `builder` is that if a `CompilationError` is found during the building steps (e.g. a type mismatch between field and constraint), this is triggers a fast failing behaviour on the message evaluator, and constraints on subsequent fields will not be built and cached.

When a `Filter` is applied to filter for a specific field descriptor, for example, the `CompilationError` on the message evaluator is not returned, but also, since constraints on a field may not be built, the constraints on the filtered field end up not being evaluated.

Some test examples demonstrating this behaviour: https://github.com/bufbuild/protovalidate-go/commit/59add0d23321dc5496b6d1cc0b45641ee2adbd04

We are currently exploring the use of `Filter` in https://github.com/bufbuild/buf for the `PROTOVALIDATE` lint rule, which would want to check for all `CompilationError`s and `Violation`s from example values set through Protovalidate options.

Some potential solutions to this:

**Option 1:** When a `CompilationError` is found during the build process for a field, set it on the field evaluator rather than the message evaluator

This would allow for all the constraints to be built, and then when a field is evaluated, then the `CompilationError` is returned. This would create a couple of behaviour changes:

- The most significant change is the fact that currently we fail fast aggressively on `CompilationError` -- we bypass any constraint evaluation. However, with `Filter`s, this behaviour could result in a confusing empty state.
- In the case where there is a `CompilationError`, we would still be building all constraints, so the evaluators built and cached would now include these constraints. This seems mostly reasonable, but it does change the contents of what is cached.
- The behaviour with fail fast is also a little different now -- in the past, `CompilationError` would fail most aggressively. With this change in place, we may evaluate a constraint, find a `Violation`, and with `FailFast` set, return the `Violation` rather than the `CompilationError` on a later field. This does not significantly degrade the usefulness of `FailFast`, it just makes `CompilationError` fail a little later.

Example of this change: https://github.com/bufbuild/protovalidate-go/commit/02a36580da789c679fceaf5ff5c6990194236cb5

**Option 2:** Fail open on all `CompilationError`s

This would be a more extreme version of option 1. It has most of the same concerns/outcomes, but would also require some consideration on how to handle failing open on `CompilationError`s. A new error type that either collects `CompilationError`s and `Violation`s, or has a `type` indicator on it would be needed.

**Option 3:** Leave this as-is and address these concerns in https://github.com/bufbuild/buf for `PROTOVALIDATE` lint rules

This is also reasonable, we can move this upstream. It is less ideal from `buf`'s perspective, since we want to use as much of `protovalidate-go`'s APIs out-of-the-box, but this is a completely viable approach.

**Solution:** The current solution is to pursue "option 1", because while it is a behaviour change, it should have minimal impact given that `pvgo` is pre 1.0 and would make the current `Validate` and `Filter` APIs usable out-of-the-box for use-cases where the caller wants to evaluate both `CompilationError`s and `Violation`s on a set of proto sources (such as the `PROTOVALIDATE` lint rule).

doriable:
Instead of setting `CompilationError`s found in the build process on the
mesasge evaluator, this sets them on their respective field evaluators.
They are then returned when the field is evaluated. This ensures that
all constraints ae built, even when a `CompilationError` is encountered.

This makes the `Filter` API easier to use because then all constraints
are built, and then `CompilationError`s are returned during the
evaluation layer.

The main change is that `CompilationError`s are no longer "fail fast"
immediately, and instead allow for all constraints to be built and
cached.

Fixes #204


doriable:
This has been validated with the upgrade to `bufbuild/buf`, https://github.com/bufbuild/buf/pull/3652

jchadwick-buf (approved):

PR #206, opened 2025-03-14
Enhance URI and IP address validation

smaye81:
This beefs up the validation for the following:

* IP address (v4 and v6)
* IP prefix (v4 and v6)
* Hostname / Port
* URI / URI reference

The validation here adheres to the defined RFC standards for each entity and passes conformance for all new conformance tests defined in https://github.com/bufbuild/protovalidate/pull/320.


cel/ipv4.go
```diff
@@ -0,0 +1,182 @@
...
+package cel
+
+import (
+	"strconv"
+)
+
+type Ipv4 struct {
+	str       string
+	index     int64
+	strLen    int64
```
timostamm, line 24: 64 bits for string length seems excessive. Can't we use whatever's typical here, probably `int`?

This applies to all parser structs.

cel/library.go
```diff
@@ -439,119 +435,48 @@ func (l library) validateEmail(addr string) bool {
...
-	case 0:
-		return true
-	case 4:
-		return address.To4() != nil
-	case 6:
-		return address.To4() == nil
-	default:
-		return false
-	}
+	return isIP(addr, ver)
```
timostamm, line 450: Instead of calling `l.validateIP(addr, 0)` in the library, why don't we call `isIp(addr, 0)` instead, and remove this method? It only adds indirection.

Same for all other similar cases. 


Somewhat related: uri.go defines functions that aren't used for URIs, for example `isIPPrefix` and `isHostAndPort`. I think we can simply move all definitions into library.go, especially if we remove the library methods.

cel/uri.go
```diff
@@ -0,0 +1,879 @@
...
+// Parses str from the current index to determine if it contains a valid
+// pchar defined as:
+//
+// pchar = unreserved / pct-encoded / sub-delims / ":" / "@".
+func (u *URI) pchar() bool {
+	return (u.unreserved() ||
+		u.pctEncoded() ||
+		u.subDelims() ||
+		u.take(':') ||
+		u.take('@'))
```
timostamm, line 724:
```suggestion
	return u.unreserved() ||
		u.pctEncoded() ||
		u.subDelims() ||
		u.take(':') ||
		u.take('@')
```

cel/uri.go
```diff
@@ -0,0 +1,879 @@
...
+// defined as:
+//
+// unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~".
+func (u *URI) unreserved() bool {
+	return (u.alpha() ||
+		u.digit() ||
+		u.take('-') ||
+		u.take('_') ||
+		u.take('.') ||
+		u.take('~'))
```
timostamm, line 790:
```suggestion
	return u.alpha() ||
		u.digit() ||
		u.take('-') ||
		u.take('_') ||
		u.take('.') ||
		u.take('~')
```

cel/uri.go
```diff
@@ -0,0 +1,879 @@
...
+		u.take('&') ||
+		u.take('\'') ||
+		u.take('(') ||
+		u.take(')') ||
+		u.take('*') ||
+		u.take('+') ||
+		u.take(',') ||
+		u.take(';') ||
+		u.take('='))
+}
```
timostamm, line 809:
```suggestion
func (u *URI) subDelims() bool {
	return u.take('!') ||
		u.take('$') ||
		u.take('&') ||
		u.take('\'') ||
		u.take('(') ||
		u.take(')') ||
		u.take('*') ||
		u.take('+') ||
		u.take(',') ||
		u.take(';') ||
		u.take('=')
}
```

cel/ipv6.go
```diff
@@ -0,0 +1,298 @@
...
+import (
+	"slices"
+	"strconv"
+)
+
+type Ipv6 struct {
+	str             string
+	index           int64
+	strLen          int64
+	pieces          []int64 // 16-bit pieces found
```
timostamm, line 26: Each piece has 16 bits, so we should use a 16-bit type here, not `int64`.

cel/uri.go
```diff
@@ -0,0 +1,879 @@
...
+	}
+	if splitIdx < 0 {
+		return !portRequired && (isHostname(str) || isIP(str, 4))
+	}
+	host := str[0:splitIdx]
+	port := str[splitIdx+1:]
+	return (isHostname(host) || isIP(host, 4)) && isPort(port)
+}
+
+// Returns true if the string is a valid port.
```
timostamm, line 162:
```suggestion
// Returns true if the string is a valid port for isHostAndPort.
```

cel/uri.go
```diff
@@ -0,0 +1,879 @@
...
+	return u.uri() || u.relativeRef()
+}
+
+// Parses str from the current index to determine if it contains a valid
+// hier-part defined as:
+//
+// hier-part = "//" authority path-abempty
+// / path-absolute
+// / path-rootless
+// / path-empty.
```
timostamm, line 221: The grammar is mangled.

cel/uri.go
```diff
@@ -0,0 +1,879 @@
...
+ *       octet, add it to the octet array
+ *     - For every other octet, add it to the octet array
+ * - Check that the octet array is valid UTF-8.
+ */
+func (u *URI) decodeURIComponent(str string) bool {
+	if _, err := url.PathUnescape(str); err != nil {
+		return false
+	}
+	return true
+}
```
timostamm, line 387: Would be possible to make it more efficient, but this is easy to read:

```go
func (u *URI) checkHostPctEncoded(str string) bool {
	unhex := func(c byte) byte {
		switch {
		case '0' <= c && c <= '9':
			return c - '0'
		case 'a' <= c && c <= 'f':
			return c - 'a' + 10
		case 'A' <= c && c <= 'F':
			return c - 'A' + 10
		}
		return 0
	}
	escaped := make([]byte, 0, len(str))
	for i := 0; i < len(str); {
		switch str[i] {
		case '%':
			escaped = append(escaped, unhex(str[i+1])<<4|unhex(str[i+2]))
			i += 3
		default:
			escaped = append(escaped, str[i])
			i++
		}
	}
	return utf8.Valid(escaped)
}
```

cel/library.go
```diff
@@ -240,7 +239,7 @@ func (l library) CompileOptions() []cel.EnvOption { //nolint:funlen,gocyclo
 					if !ok {
 						return types.Bool(false)
 					}
-					return types.Bool(l.validateURI(s, true))
+					return types.Bool(l.validateURI(s))
```
smaye81, line 242: Not usre if you'd rather these also be inlined as `types.Bool(newURI(s).uri()`, etc. Lmk if so.

timostamm: I'd prefer a `func isURI`, with the same doc comments as the source implementation. This makes porting easier because the `library` struct is out of the picture.

Alternatively, we can make all functions methods of `library`, but I don't see the point. They don't use state, and I think it's good that they don't.

Okay, `isEmail` uses state. But it doesn't have to:

```go
// See https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
var emailRegexp = regexp.MustCompile("...")

// helpful comments
func isEmail(addr string) bool {
	return emailRegexp.MatchString(addr)
}
``` 

Whichever we do, it should be consistent.

cel/library.go
```diff
@@ -439,135 +438,1241 @@ func (l library) validateEmail(addr string) bool {
...
+		strLen:        len(str),
+		pieces:        make([]uint16, 0),
+		doubleColonAt: -1,
+		dottedAddr:    nil,
+	}
+}
+
+/**
+ * Returns true if the string is an IPv4 or IPv6 address, optionally limited to
+ * a specific version.
```
timostamm, line 897: I'm not sure I've ever seen block comments in Go code, and would have expected line comments here. Deferring to @rodaine, @jchadwick-buf.

cel/library.go
```diff
@@ -439,135 +438,1241 @@ func (l library) validateEmail(addr string) bool {
...
+}
+
+// newIpv6 creates a new ipv6 based on str.
+func newIpv6(str string) *ipv6 {
+	return &ipv6{
+		str:           str,
+		strLen:        len(str),
+		pieces:        make([]uint16, 0),
+		doubleColonAt: -1,
+		dottedAddr:    nil,
```
timostamm, line 891: Can't we remove these constructor functions?

- Since go strings are immutable, I suspect that storing `strLen` doesn't give us any benefit over `len(i.str)`.
- I'm pretty we can just let `pieces` be a nil slice.
- We have a separate bool to track whether we've seen a double colon, so `doubleColonAt` doesn't need to be initialized to `-1`. 
- `dottedAddr: nil` is redundant.

This only leaves `str`. Which means it's feasible to use the struct without a constructor function:

```diff
-		dotted := newIpv4(i.dottedRaw)
+		dotted := &ipv4{str: i.dottedRaw}
```

@jchadwick-buf, @rodaine WDYT?

timostamm (commented):
Looks right to me, just left three comments.

cel/library.go
rodaine: Line comments only please.

cel/library.go
rodaine: `strLen` is indeed unnecessary, and `pieces` should be left as a nil slice. The only one I'd still set is `doubleColonoAtr` to avoid having invalid state (even if there is an external boolean tracking the colon's presence).

cel/library.go
smaye81: The only problem is that the linter complains about invalid formatting if you try to align things according to the RFC grammar with line comments (see `hierPart` for an example). 

What is the Go standard practice for getting around that @rodaine? Other than adding an ignore I guess.

rodaine: Check out the doc comment (er...) [docs on code blocks](https://tip.golang.org/doc/comment#code).

rodaine: i'm gonna guess it's the tabbing on 1142. Does running `golangci-lint run --fix` change it?

smaye81: Yeah it mangles it even further. The best I can do it seems to have a line break in between. Go seems to really want that to happen so 🤷

cel/library.go
smaye81: I fixed up the initialization and removed the constructor function for `NewUri`, but I left the other two constructor functions. I can remove them if you feel strongly about it, but I just think the code is much cleaner and reads much nicer with them for the IPv4 and IPv6 addresses. Plus, I'm not sure it really buys us anything removing them other than a few LOC?

cel/library.go
```diff
@@ -429,145 +424,1232 @@ func (l library) uniqueBytes(list traits.Lister) ref.Val {
...
 	}
+	return true
 }
 
-func (l library) validateURI(val string, checkAbs bool) bool {
-	uri, err := url.Parse(val)
-	if err != nil {
+// relative-part = "//" authority path-abempty.
+//
+//				 / path-absolute
```
smaye81, line 1121: Note this looks fine on my machine 🤷

cel/library.go
timostamm: Fixing the formatting: https://github.com/bufbuild/protovalidate-go/pull/208

cel/library.go
timostamm: :+1: I'll die on another hill.

timostamm (approved):
Looks right to me.

Can we reference https://github.com/bufbuild/protovalidate/pull/320 in the description for context?

smaye81:
@rodaine cool to merge this into my `upgrade_conformance` branch?

rodaine (approved):

PR #207, opened 2025-03-18
Unwrap redundant parentheses

timostamm:


smaye81 (approved):
Bad merge. I had some of these implemented, but they got removed.

PR #217, opened 2025-04-14
Fix panic in isUri

timostamm:
https://github.com/bufbuild/protovalidate-go/pull/217/commits/ccafab4f2a95294b2d32dbb1c00df11edf51cecb improves the documentation, and https://github.com/bufbuild/protovalidate-go/pull/217/commits/e7c14c9b852bc86bf86546818d282c306b08b2b6 avoids a lint ignore. 

The panic is fixed in https://github.com/bufbuild/protovalidate-go/pull/217/commits/e7c14c9b852bc86bf86546818d282c306b08b2b6.

### DIFF
--- a/cel/library.go
+++ b/cel/library.go
@@ -425,17 +425,20 @@ func (l library) uniqueBytes(list traits.Lister) ref.Val {
 	return types.Bool(true)
 }
 
-// isEmail returns true if addr is a valid email address.
+// isEmail reports whether val an email address, for example "foo@example.com".
 //
-// This regex conforms to the definition for a valid email address from the HTML standard.
+// Conforms to the definition for a valid email address from the HTML standard.
 // Note that this standard willfully deviates from RFC 5322, which allows many
 // unexpected forms of email addresses and will easily match a typographical
 // error.
-func isEmail(addr string) bool {
-	return emailRegex.MatchString(addr)
+func isEmail(val string) bool {
+	return emailRegex.MatchString(val)
 }
 
-// isURI validates whether val is a valid URI.
+// isURI reports whether val is a URI, for example "https://example.com/foo/bar?baz=quux#frag".
+//
+// URI is defined in the internet standard RFC 3986.
+// Zone Identifiers in IPv6 address literals are supported (RFC 6874).
 func isURI(val string) bool {
 	uri := &uri{
 		str: val,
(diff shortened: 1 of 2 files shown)


cel/library.go
```diff
@@ -1179,7 +1186,7 @@ func (u *uri) scheme() bool {
 				break
 			}
 		}
-		if u.str[u.index] == ':' {
+		if u.peek(':') {
```
timostamm, line 1189: This fixes the panic for the input "A://". I suggest we adopt `peek` to other implementations where necessary.

cel/library.go
```diff
@@ -1282,15 +1289,12 @@ func (u *uri) checkHostPctEncoded(str string) bool {
 
 // host parses the rule:
 //
-//	host = IP-literal / IPv4address / reg-name.
+//	host = IP-literal / IPv4address / reg-name
 func (u *uri) host() bool {
-	if u.index >= len(u.str) {
-		return false
-	}
```
timostamm, line 1289: This was returning too early. We would never enter regName, which can be empty.

cel/library.go
```diff
@@ -425,25 +425,34 @@ func (l library) uniqueBytes(list traits.Lister) ref.Val {
 	return types.Bool(true)
 }
 
-// isEmail returns true if addr is a valid email address.
+// isEmail reports whether val an email address, for example "foo@example.com".
```
rodaine, line 428:
```suggestion
// isEmail reports whether val is an email address, for example "foo@example.com".
```

cel/library.go
```diff
@@ -1746,3 +1750,12 @@ func (u *uri) take(char byte) bool {
 	}
 	return false
 }
+
+func (u *uri) takeDoubleSlash() bool {
+	first := u.take('/')
+	return first && u.take('/')
```
rodaine, line 1756: Go [short-circuits binary expressions](https://go.dev/play/p/X4YSB1FsTaU):

```suggestion
	return u.take('/') && u.take('/')
```

cel/library.go
timostamm: Thanks!

cel/library.go
timostamm: `u.take('/') && u.take('/')` is incorrectly flagged by [SA4000](https://staticcheck.dev/docs/checks/#SA4000).

rodaine: Ah go ahead and add a nolint directive. Should look like `//nolint:staticcheck // false positive of SA4000

PR #221, opened 2025-04-16
Fix Ignore Always detection

smaye81:
This fixes a small bug in how an `IGNORE_ALWAYS` annotation is processed.

Previously, when `IGNORE_ALWAYS` was detected, the builder just skipped all the `processXXX` methods for a value. However, the field's `EvaluateMessage` was still being called and there was no check present to see if we should skip validation. Instead, it would continue evaluating and since there was no value present, it would fail validation.

This adds a new `IgnoreAlways` attribute on the `field` struct so that at evaluation time, we can determine whether everything should be skipped entirely. In addition, it renames `shouldSkip` to `shouldIgnoreAlways` for consistency.

Note: tested this against Protovalidate 0.10.4 and can confirm that all conformance tests pass.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -213,14 +213,11 @@ func (bldr *builder) buildField(
 		Value: value{
 			Descriptor: fieldDescriptor,
 		},
-		Required:     fieldConstraints.GetRequired(),
-		IgnoreAlways: bldr.shouldIgnoreAlways(fieldConstraints),
-		IgnoreEmpty: fieldDescriptor.HasPresence() ||
-			bldr.shouldIgnoreEmpty(fieldConstraints),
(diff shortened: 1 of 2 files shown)


smaye81:
@rodaine lmk know if https://github.com/bufbuild/protovalidate-go/pull/221/commits/813e2f38ba4f3ed822dfcfd971652a5818cd2737 is what you had in mind. It was easy to collapse to just the enum for fields, but values are a bit trickier. 

I didn't want to go on a refactoring spree since `value` only uses the one ignore attribute, but I don't really love how fields have ignore methods, but the value doesn't. So, lmk what you think.

PR #223, opened 2025-04-18
Relax go version requirement in go.mod

pkwarren:
Update go.mod to relax version requirement to 1.23.0. There is nothing specific in later patch releases that is required by this library, so we should not require consumers to upgrade.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/bufbuild/protovalidate-go
 
-go 1.23.4
+go 1.23.0
 
 require (
 	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.36.6-20250307204501-0409229c3780.1


pkwarren:
Using a later patch version is causing downstream dependabot builds to add a `toolchain` and also fail when `go mod tidy` further changes the go.mod file: https://github.com/bufbuild/protoschema-plugins/pull/68

Running `go mod tidy` on that PR locally further adds this change:
```diff
@@ -1,7 +1,8 @@
 module github.com/bufbuild/protoschema-plugins
 
-go 1.23.0
-toolchain go1.24.1
+go 1.23.4
+
+toolchain go1.24.2
 
 require (
 	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.36.6-20250307204501-0409229c3780.1
```

timostamm (approved):

PR #225, opened 2025-04-22
Implement getField CEL function

jchadwick-buf:
I think it has to be a global overload, a member overload would require a generic on the receiver which doesn't seem possible.

I'm proposing this as an eventual replacement (before 1.0) of our hack around the fact that the `in` identifier is reserved in CEL. This is especially urgent for protovalidate-cc which is currently carrying patches to the CEL implementation in order to enable it, since cel-cpp doesn't allow this sort of functionality to be added in at runtime.

### DIFF
--- a/cel/library.go
+++ b/cel/library.go
@@ -30,6 +30,7 @@ import (
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
 	"github.com/google/cel-go/ext"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/reflect/protoregistry"
 	"google.golang.org/protobuf/types/dynamicpb"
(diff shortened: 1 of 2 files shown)


cel/library.go
```diff
@@ -80,20 +79,15 @@ func (l library) CompileOptions() []cel.EnvOption { //nolint:funlen,gocyclo
 				[]*cel.Type{cel.AnyType, cel.StringType},
 				cel.AnyType,
 				cel.FunctionBinding(func(values ...ref.Val) ref.Val {
-					message, ok := values[0].Value().(proto.Message)
+					message, ok := values[0].(interface{ Get(index ref.Val) ref.Val })
```
rodaine, line 82: Could use `traits.Indexer`?

rodaine:
Also, re:

> a member overload would require a generic on the receiver which doesn't seem possible.

If you make the receiver type dyn, I think it would work?

jchadwick-buf:
Oops, I saw this after me and Nick moved to merge the protovalidate PR. I've been pondering but I think it's not worth losing too much sleep over; we're still pre-1.0 so we can always try to make this prettier later, especially since the protovalidate protobuf descriptors used for standard constraints are always strongly tied to the implementation version.

cel/library.go
jchadwick-buf: lol, yeah. Fixed

rodaine (approved):

PR #227, opened 2025-04-23
Upgrade to Protovalidate v0.11.0

smaye81:
This upgrades to the latest v0.11.0 release of Protovalidate. As a result, it conforms to the convention of using `rule` everywhere instead of `constraint`. All usages of `constraint` have been removed.

In addition, this adds a few additional tests to the expected_failures list as a result of v0.11.0.

Also, there are two failing unit tests as a result of the upgrade, which will be fixed in a follow-up as they seem to be related to the failing conformance tests. These failing unit tests are marked with a `TODO`.

### DIFF
--- a/.github/workflows/buf.yaml
+++ b/.github/workflows/buf.yaml
@@ -28,8 +28,6 @@ jobs:
           token: ${{ secrets.BUF_TOKEN }}
           format: true
           lint: true
-          # Temp disable breaking checks due to upgrade to v0.11.0
-          breaking: false
           push: false
           archive: false
       - name: Check Generate
--- a/.github/workflows/buf.yaml
+++ b/.github/workflows/buf.yaml
@@ -28,6 +28,8 @@ jobs:
           token: ${{ secrets.BUF_TOKEN }}
           format: true
           lint: true
+          # Temp disable breaking checks due to upgrade to v0.11.0
+          breaking: false
           push: false
           archive: false
       - name: Check Generate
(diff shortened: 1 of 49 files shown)


.github/workflows/buf.yaml
```diff
@@ -28,6 +28,8 @@ jobs:
           token: ${{ secrets.BUF_TOKEN }}
           format: true
           lint: true
+          # Temp disable breaking checks due to upgrade to v0.11.0
+          breaking: false
```
nicksnyder, line 32: You can revert this and just add the `Buf skip breaking` label (which I added to the PR)
```suggestion
```

Makefile
```diff
@@ -14,9 +14,9 @@ GO ?= go
 ARGS ?= --strict_message --strict_error
 GOLANGCI_LINT_VERSION ?= v1.64.8
 # Set to use a different version of protovalidate-conformance.
-# Should be kept in sync with the version referenced in proto/buf.lock and
+# Should be kept in sync with the version referenced in buf.yaml and
 # 'buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go' in go.mod.
```
nicksnyder, line 18: I am wondering if we can/should advertise on the main readme the conformance version this library is targeting

smaye81: I like that idea

timostamm: Would definitely be helpful. Perhaps even more so to have it in the release notes.

nicksnyder (approved):

PR #229, opened 2025-04-24
Fix failures

smaye81:
This fixes all failing conformance tests with two fixes:

* The `getField` implementation should be defined with `cel.DynType` so that the type is determined at runtime. Attempting to use an `AnyType` will cause issues when using `getField` in a string `format` statement.
* Updating to latest cel-go, which fixes some issues matching stringified values.


jchadwick-buf (approved):
Oopsies.

jchadwick-buf (approved):

PR #235, opened 2025-04-29
Rewrite import path to buf.build/go/protovalidate

jchadwick-buf:
Should be straight-forward enough. You can validate that this works by running something like:

```shell
$ go run buf.build/go/protovalidate/internal/cmd/protovalidate-conformance-go@90995783ab7ab3fc0c67f3336349eaae3aac8ed8 </dev/null
```

...since Go can see branches inside of PRs.


rodaine:
We will want to do another release of the GH-path to include a redirection message to the new import path to help folks migrate.

jchadwick-buf:
Got it. Released v0.10.1 with the notice.

rodaine (approved):

PR #202, opened 2025-03-11
Upgrade validation logic

smaye81:
This includes upgrades to validation logic for:

- URIs, URI references
- IP addresses and prefixes (v4 and v6)
- Hostname, Port
- Email Address

The validation here adheres to the defined RFC standards for each entity and passes conformance for all new conformance tests defined in:

- https://github.com/bufbuild/protovalidate/pull/320
- https://github.com/bufbuild/protovalidate/pull/341

PR #242, opened 2025-04-30
Enable staticcheck QF1006 linter

pkwarren:
Enable the linter which checks for if conditions that could be lifted into loop conditions. This removes nesting and makes the code easier to read (in my opinion).

### DIFF
--- a/cel/library.go
+++ b/cel/library.go
@@ -543,10 +543,7 @@ func (i *ipv4) addressPrefix() bool {
 // prefixLength parses the length of the prefix and stores the value in prefixLen.
 func (i *ipv4) prefixLength() bool {
 	start := i.index
-	for {
-		if i.index >= len(i.str) || !i.digit() {
-			break
-		}
+	for i.digit() {
 		if i.index-start > 2 {
 			// max prefix-length is 32 bits, so anything more than 2 digits is invalid
 			return false
@@ -593,10 +590,7 @@ func (i *ipv4) addressPart() bool {
 // decOctet parses str from the current index to determine a decimal octet.
 func (i *ipv4) decOctet() bool {
 	start := i.index
-	for {
-		if i.index >= len(i.str) || !i.digit() {
-			break
-		}
+	for i.digit() {
 		if i.index-start > 3 {
 			// decimal octet can be three characters at most
 			return false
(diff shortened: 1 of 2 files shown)


cel/library.go
```diff
@@ -626,6 +620,9 @@ func (i *ipv4) decOctet() bool {
 //
 //	DIGIT = %x30-39  ; 0-9
 func (i *ipv4) digit() bool {
+	if i.index >= len(i.str) {
```
pkwarren, line 623: The `digit()` helpers were inconsistent - the URI one had this guard in place but others didn't (and thus had to guard in each loop where they were used). This feels better to simplify the call sites.

cel/library.go
```diff
@@ -842,11 +830,8 @@ func (i *ipv6) zoneID() bool {
 func (i *ipv6) dotted() bool {
 	start := i.index
 	i.dottedRaw = ""
-	for {
-		if i.index < len(i.str) && (i.digit() || i.take('.')) {
-			continue
-		}
-		break
+	for i.digit() || i.take('.') {
+		// Consume '*( DIGIT "." )'
```
pkwarren, line 834: This is one place where we're a bit loose w/ the parsing - basically parsing any 7+ length sequence of digits / '.' into a potential IP address. In other places in the parser we're a bit more strict where we don't advance the parser's position if we don't match the expected sequences.

I don't think this is an issue in practice, but potentially a place where we could make it consistent.

rodaine (approved):

cel/library.go
timostamm: Agreed, consuming a DIGIT should simply return false if we're at the end of the input, and not require guards on call-sites.

cel/library.go
timostamm: Potentially, IPv6 and IPv4 parsers could be merged, which would also remove the extra allocation https://github.com/bufbuild/protovalidate-go/pull/202#issuecomment-2785946148.

cel/library.go
```diff
@@ -1608,13 +1573,8 @@ func (u *uri) segmentNz() bool {
...
 	start := u.index
-	for {
-		if !u.unreserved() &&
-			!u.pctEncoded() &&
-			!u.subDelims() &&
-			!u.take('@') {
-			break
-		}
+	for u.unreserved() || u.pctEncoded() || u.subDelims() || u.take('@') {
+		// Consume '*( unreserved / pct-encoded / sub-delims / "@" )'
```
timostamm, line 1577: ❤️

timostamm (approved):

cel/library.go
jchadwick-buf: Merging the IPv4 and IPv6 parsers make sense to me; in protovalidate-cc where I didn't follow the TS implementation exactly I did share some of the logic between them anyway.

PR #247, opened 2025-05-13
Correctly validate WKT rules on incorrect field types

smaye81:
A bug existed in Protovalidate-Go where it was possible to add a WKT rule to an incorrect field type. For example, given the following:

```proto
message RulesWrongTypeExample {
  message WrongType {
    string name = 1;
  }
  WrongType f = 1 [(buf.validate.field).timestamp.lt = {
    seconds: 1
    nanos: 1
  }];
}
```

the expected behavior would be that a compilation error would be raised due to applying a WKT Timestamp rule to a non-Timestamp field. Note that Buf Lint catches this scenario, but if lint errors are ignored, this would go through validation with no issues.

This changes the behavior to throw a compilation error if no rule descriptor is found and the field is a message type. Since the only valid rule descriptor for a message field is for well-known types, if we don't find a descriptor for a message field, we know that this message is using a WKT rule on a non-WKT field.

There was also an issue where `processStandardRules` was being called twice for wrapper fields, once for the wrapper and once for the wrapped value. This adds a short-circuit return to that function so that it simply ignores processing standard rules for the wrapper.


smaye81:
@rodaine I don't love this solution, but I don't know if there's an easier way without refactoring the internals here. I tested with additional conformance tests locally and this passes all of them successfully, correctly identifying rules on incorrect scalar, message, and wrapper types.

If you see an easier way to do this, lmk. This one was a bit tricky to resolve.

rodaine (approved):

PR #258, opened 2025-06-13
Add implicit `IGNORE_IF_UNPOPULATED` for `MessageOneofRule` fields

srikrsna-buf:


builder.go
```diff
@@ -250,15 +251,20 @@ func (bldr *builder) processFields(
 func (bldr *builder) buildField(
 	fieldDescriptor protoreflect.FieldDescriptor,
 	fieldRules *validate.FieldRules,
+	msgRules *validate.MessageRules,
 	cache messageCache,
 ) (field, error) {
+	ignore := fieldRules.GetIgnore()
+	if !fieldRules.HasIgnore() && isPartOfMessageOneof(msgRules, fieldDescriptor) {
```
timostamm, line 258: Good call to only use IGNORE_IF_UNPOPULATED if `FieldRules.ignore` is unset. 

We give this example for IGNORE_ALWAYS in validate.proto:

```proto
message MyMessage {
The field's rules will always be ignored, including any validation's
on value's fields.
  MyOtherMessage value = 1 [
    (buf.validate.field).ignore = IGNORE_ALWAYS];
}
```

This should continue to work with a `MesssageRules.oneof`.

We should have conformance tests for this behavior.

timostamm (approved):

PR #262, opened 2025-06-13
Delete unused github-tag-action/entrypoint.sh

timostamm:


timostamm:
I don't see this referenced anywhere. Can we delete it, @rodaine?

rodaine (approved):
Yep! I think it's vestigial.

PR #292, opened 2025-11-26
Fix field paths for groups

timostamm:
With https://github.com/bufbuild/protovalidate/pull/439, we're supporting proto2 Groups and Editions delimited message fields.

Unfortunately, the logic to create the field paths for violations does not handle groups correctly: For a `group Optional`, we create a `buf.validate.FieldPathElement` with `field_name` "Optional". The correct field name is "optional".

Also see https://github.com/bufbuild/protovalidate/pull/443


error_utils_test.go
```diff
@@ -19,11 +19,47 @@ import (
...
+		require.Equal(t, descriptorpb.FieldDescriptorProto_TYPE_GROUP, element.GetFieldType())
+	})
+	t.Run("message field", func(t *testing.T) {
+		t.Parallel()
+
+		messageFieldDescriptor := (&descriptorpb.FileDescriptorProto{}).ProtoReflect().Descriptor().Fields().ByName("options")
+		element := fieldPathElement(messageFieldDescriptor)
+		require.NotNil(t, element)
+		require.Equal(t, int32(8), element.GetFieldNumber())
+		require.Equal(t, "options", element.GetFieldName())
```
timostamm: Without the fix to error_utils.go, this would yield:

```
                Error:          Not equal: 
                                expected: "optional"
                                actual  : "Optional"
```

rodaine (approved):

PR #295, opened 2025-12-16
Add ability to depend on an arbitrary upstream commit

srikrsna-buf:


Makefile
```diff
@@ -15,10 +15,20 @@ LICENSE_IGNORE := -e internal/testdata/
...
+PROTOVALIDATE_PROTO_PATH := buf.build/bufbuild/protovalidate:$(PROTOVALIDATE_VERSION)
+PROTOVALIDATE_TESTING_PROTO_PATH := buf.build/bufbuild/protovalidate-testing:$(PROTOVALIDATE_VERSION)
+ifneq ($(shell echo ${PROTOVALIDATE_VERSION} | grep -E "^v\d+\.\d+.\d+(-.+)?$$"), $(PROTOVALIDATE_VERSION))
+  PROTOVALIDATE_PROTO_PATH = https://github.com/bufbuild/protovalidate.git\#subdir=proto/protovalidate,ref=$(PROTOVALIDATE_VERSION)
+  PROTOVALIDATE_TESTING_PROTO_PATH = https://github.com/bufbuild/protovalidate.git\#subdir=proto/protovalidate-testing,ref=$(PROTOVALIDATE_VERSION)
+  PROTOVALIDATE_RELEASE_VERSION = FALSE
+else
+  PROTOBUF_GO_VERSION := $(shell go list -json -m google.golang.org/protobuf  | jq -r '.Version')
+  PROTOVALIDATE_GEN_SDK_VERSION := $(shell buf registry sdk version --module=buf.build/bufbuild/protovalidate:$(PROTOVALIDATE_VERSION) --plugin=buf.build/protocolbuffers/go:$(PROTOBUF_GO_VERSION))
+endif
```
hudlow: I would find this clearer, especially because it matches up with the order of the if/else in the `upstream` target below.

```suggestion
# Set to use a different version of protovalidate
PROTOVALIDATE_VERSION ?= 895eefca6d1346f742fc18b9983d40478820906d

ifeq ($(shell echo ${PROTOVALIDATE_VERSION} | grep -E "^v\d+\.\d+.\d+(-.+)?$$"), $(PROTOVALIDATE_VERSION))
	PROTOVALIDATE_RELEASE_VERSION := TRUE

	PROTOVALIDATE_PROTO_PATH := buf.build/bufbuild/protovalidate:$(PROTOVALIDATE_VERSION)
	PROTOVALIDATE_TESTING_PROTO_PATH := buf.build/bufbuild/protovalidate-testing:$(PROTOVALIDATE_VERSION)
  PROTOBUF_GO_VERSION := $(shell go list -json -m google.golang.org/protobuf  | jq -r '.Version')
  PROTOVALIDATE_GEN_SDK_VERSION := $(shell buf registry sdk version --module=buf.build/bufbuild/protovalidate:$(PROTOVALIDATE_VERSION) --plugin=buf.build/protocolbuffers/go:$(PROTOBUF_GO_VERSION))
else
  PROTOVALIDATE_RELEASE_VERSION := FALSE

  PROTOVALIDATE_TESTING_PROTO_PATH := https://github.com/bufbuild/protovalidate.git\#subdir=proto/protovalidate-testing,ref=$(PROTOVALIDATE_VERSION)
  PROTOVALIDATE_PROTO_PATH := https://github.com/bufbuild/protovalidate.git\#subdir=proto/protovalidate,ref=$(PROTOVALIDATE_VERSION)
endif
```

Makefile
```diff
@@ -15,10 +15,20 @@ LICENSE_IGNORE := -e internal/testdata/
 GO ?= go
 ARGS ?= --strict_message --strict_error
 GOLANGCI_LINT_VERSION ?= v2.4.0
-# Set to use a different version of protovalidate-conformance.
-# Should be kept in sync with the version referenced in buf.yaml and
-# 'buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go' in go.mod.
-CONFORMANCE_VERSION ?= v1.1.0
```
hudlow: Can we add a `target` like this...
```
.PHONY: assert-release-version
assert-release-version:
	@# Used in PRs to main to make sure we don't merge a non-release version.
	test $(PROTOVALIDATE_RELEASE_VERSION) = TRUE
```

And then add a workflow triggered on PRs to `main` that runs it?

srikrsna-buf: We intend to merge this as is i.e. always use latest main. Doing it all in one PR will not allow us to work on individual updates to protovalidate.

Makefile
srikrsna-buf: Simplified in https://github.com/bufbuild/protovalidate-go/pull/295/commits/cd9c8b4e6ec246813aaf855203dfc0bd6c5dda28

hudlow (approved):

rodaine (changes requested):
I know this makes development of pv-go easier, but we don't want users to rely on this generated code and add to our support burden when things don't align properly. We should be able to do this in a transparent way.

srikrsna-buf:
> but we don't want users to rely on this generated code and add to our support burden when things don't align properly

They won't, because to cut a release we will use a release version of protovalidate just like today and that removes the replace directive and the generated code. The only problem I see is if users depend on an arbitrary commit which I don't think we need to support.

jhump:
There's no PR description that states the problem this solves. Is this just so we can development and test against arbitrary versions of the protos? I would think we could include a lot of that plumbing, but then also have some other CI check to make sure we don't ever commit that to main. I think commits on main should always use a release of the protos.

What kinds of features are we thinking of that might necessitate working on unreleased versions of the proto on the main branch? I imagine they could be potentially be done via longer-lived dev branches instead of merging such a dependency to main, no?

rodaine:
Closing for now until we have a chance to revisit this.

PR #310, opened 2026-03-10
Add ability to depend on arbitrary upstream commits

srikrsna-buf:
With https://github.com/bufbuild/protovalidate/pull/470, all commits of protovalidate and protovalidate-testing module will be available on the BSR. This means we can depend on the generated SDK versions of those commits. 

This PR adds the ability to depend on an arbitrary upstream commit. This drastically improves the development cycle by letting us review and merge PRs for each individual change in `validate.proto`, rather than bundling all changes from the previous release into a single PR.

It uses the git commit/tag/branch as the pointer because there are two different BSR modules, and each will have its own BSR commit name. 

To avoid merging zeroed versions of the generated SDK, this ensures that only release versions or commits on the main version can be merged. The CI failure is because of this. Once https://github.com/bufbuild/protovalidate/pull/470 is merged and we update the reference here, it should pass. 

Because of the version update, there were some errors in the example test. I fixed them in a separate commit.

### DIFF
--- a/.github/workflows/check-gen-dep.yaml
+++ b/.github/workflows/check-gen-dep.yaml
@@ -1,0 +1,25 @@
+name: Check generated SDK dependency
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    name: Check gen dependency version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+      - name: Check for zeroed-out version
+        run: |
+          if grep -q 'buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go.*00000000000000' go.mod; then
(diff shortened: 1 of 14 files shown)


.github/workflows/check-gen-dep.yaml
```diff
@@ -0,0 +1,25 @@
+name: Check generated SDK dependency
```
emcfarlane, line 1: I like this idea. One suggestion: instead of `exit 1` (which shows a red X and looks like broken CI), consider using the GitHub Checks API with an `action_required` conclusion. This still blocks merging when set as a required check, but shows as a distinct warning rather than a failure. Making it clear this is "waiting on upstream" not "something is broken."

Rough sketch:

```yaml
- name: Check for zeroed-out version
  uses: actions/github-script@v7
  with:
    script: |
      const gomod = require('fs').readFileSync('go.mod', 'utf8');
      const zeroed = /buf\.build\/gen\/go\/bufbuild\/protovalidate\/protocolbuffers\/go.*00000000000000/.test(gomod);
      if (zeroed) {
        await github.rest.checks.create({
          owner: context.repo.owner,
          repo: context.repo.repo,
          name: 'Upstream dependency check',
          head_sha: context.sha,
          status: 'completed',
          conclusion: 'action_required',
          output: {
            title: 'Waiting for upstream protovalidate merge',
            summary: '`go.mod` has a zeroed-out pseudo-version. Update `PROTOVALIDATE_VERSION` once the upstream PR is merged and re-run `make sync-upstream`.'
          },
          details_url: 'https://github.com/bufbuild/protovalidate/pulls'
        });
      }
```

rodaine:
Closing for now until we have a chance to revisit this.

PR #314, opened 2026-04-16
Bump protovalidate to `1.2.0`

srikrsna-buf:
Bumps protovalidate to `1.2.0`. This builds on #313.


stefanvanburen (approved):

stefanvanburen:
actually, silly question, but: do we need to implement https://github.com/bufbuild/protovalidate/pull/465 here?

srikrsna-buf:
> actually, silly question, but: do we need to implement [bufbuild/protovalidate#465](https://github.com/bufbuild/protovalidate/pull/465) here?

Because it is CEL and we use reflection to figure out the rules, we don't need to. We will need to make changes if we add a predefined rules for a new type like the recent `FieldMaskRules`

PR #315, opened 2026-04-16
Add release.yml

AdrienVannson:
This adds a `release.yml` file to remove dependabot PRs from auto-generated release notes.


.github/release.yml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  exclude:
+    labels:
+      - ignore-for-release
+    authors:
+      - dependabot[bot]
+      - renovate[bot]
```
srikrsna-buf: We don't need renovate here
```suggestion
```

AdrienVannson: In this issue, Timo mentioned:

> python repositories use renovate, not dependabot. Would be smart to exclude renovate as well from the start, for every repo.

(I don't have a strong preference on any option, it's easy to add renovate later if we want to but it doesn't harm to include it now)

srikrsna-buf: I am not sure if he meant for every repo or every python repo. This is a go repo and renovate will be unfamiliar to go developers and keeping it will only confuse them.

AdrienVannson: Renovate supports go too, not only python. But if you prefer, let's remove it now, it is easy to add back in the future if we need to

srikrsna-buf (approved):

PR #318, opened 2026-04-29
Update Makefile to have a variable for license-header version, a variable for exclusions for license headers, and update the task for generating license headers to respect the variable

jonbodner-buf:


Makefile
```diff
@@ -77,10 +79,13 @@ generate-proto: $(BIN)/buf
 
 .PHONY: generate-license
 generate-license: $(BIN)/license-header
-	$(BIN)/license-header \
-		--license-type apache \
-		--copyright-holder "Buf Technologies, Inc." \
-		--year-range "$(COPYRIGHT_YEARS)"
+	comm -23 \
+		<(git ls-files --cached --modified --others --no-empty-directory --exclude-standard | sort -u | grep -v $(LICENSE_IGNORE) ) \
+		<(git ls-files --deleted | sort -u) | \
```
timostamm: Are you sure that git ls-files is needed?

jonbodner-buf: I'm not, but this was copied from pv-cc. The previous target didn't do any exclusions.

timostamm: The target in pv-cc hasn't been updated in 3 years, but this one here was recently updated. I dropped the `git ls-files` part because the license-header command already does it, see https://github.com/bufbuild/buf/blob/v1.67.0/private/pkg/licenseheader/cmd/license-header/main.go#L204 and https://github.com/bufbuild/buf/blob/main/private/pkg/git/git.go#L145.

The command didn't always support this, and I guess we're still paying off the debt.

jonbodner-buf: Ah! I can simplify then. Thanks!

Makefile
```diff
@@ -18,6 +18,8 @@ GOLANGCI_LINT_VERSION ?= v2.9.0
 # Should be kept in sync with the version referenced in buf.yaml and
 # 'buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go' in go.mod.
 CONFORMANCE_VERSION ?= v1.2.0
+LICENSE_IGNORE := -e .github/ -e .golangci.yml -e buf.gen.yaml -e buf.yaml -e conformance/expected_failures.yaml
```
timostamm, line 21: I don't believe this is needed. The command only touches files with the file extensions defined here: https://github.com/bufbuild/buf/blob/v1.67.0/private/pkg/licenseheader/licenseheader.go#L75

jonbodner-buf: It was updated to include YAML files in 

https://github.com/bufbuild/buf/blob/v1.69.0/private/pkg/licenseheader/licenseheader.go#L92

which is what the LICENSE_HEADER_VERSION is set to.

timostamm (approved):

PR #324, opened 2026-06-15
Set field descriptor on map entry violations

### LINKED ISSUE #322: [BUG] `Violation.FieldDescriptor` is `nil` for `map.values` violations
## Description

`FieldDescriptor` is `nil` on violations for map values, which is problematic when attempting to handle violations by referring to the source field's protoreflect type.

## Steps to Reproduce

https://github.com/haines/protovalidate-violation-field-descriptor

```proto
syntax = "proto3";

package example.v1;

import "buf/validate/validate.proto";

option go_package = "github.com/haines/protovalidate-violation-field-descriptor/gen/example/v1;examplev1";

enum Value {
  VALUE_UNSPECIFIED = 0;
  VALUE_A = 1;
  VALUE_B = 2;
}

message Example {
  Value value = 1 [(buf.validate.field).enum = {
    in: [
      1,
      2
    ]
  }];

  map<string, Value> values = 2 [(buf.validate.field).map.values.enum = {
    in: [
      1,
      2
    ]
  }];
}
```

```go
package main

import (
	"errors"
	"fmt"
	"log"

	"buf.build/go/protovalidate"
	examplev1 "github.com/haines/protovalidate-violation-field-descriptor/gen/example/v1"
)

func main() {
	err := protovalidate.Validate(&examplev1.Example{
		Value: examplev1.Value_VALUE_UNSPECIFIED,
		Values: map[string]examplev1.Value{"foo": examplev1.Value_VALUE_UNSPECIFIED},
	})
	validationErr, ok := errors.AsType[*protovalidate.ValidationError](err)
	if !ok {
		log.Fatalln(err)
	}

	for _, violation := range validationErr.Violations {
		fmt.Println(violation)
		fmt.Println(violation.FieldDescriptor)
		fmt.Println()
	}
}
```

```console
$ go run github.com/haines/protovalidate-violation-field-descriptor@v0.0.0-20260615182102-eacba7874c96
value: must be in list [1, 2]
FieldDescriptor{Syntax: proto3, FullName: example.v1.Example.value, Number: 1, Cardinality: optional, Kind: enum, HasJSONName: true, JSONName: "value", Enum: example.v1.Value}

values["foo"]: must be in list [1, 2]
<nil>
```

## Expected Behavior

`violation.FieldDescriptor` is set to the map values' field descriptor (i.e. what you get by calling `MapValue()` on the parent field descriptor).

## Actual Behavior

`violation.FieldDescriptor` is `nil`.

## Screenshots/Logs

N/A

## Environment

- **Operating System**: N/A
- **Version**: N/A
- **Compiler/Toolchain**: N/A
- **Protobuf Compiler & Version**: buf v1.70.0, buf.build/protocolbuffers/go:v1.36.11
- **Protovalidate Version**: v1.2.0

## Additional Context

I am trying to work around #125 by creating a new error message for violations of `enum.in` rules.

I came up with the following

```go
for _, violation := range validationErr.Violations {
	if violation.RuleDescriptor.FullName() == "buf.validate.EnumRules.in" {
		allValues := violation.FieldDescriptor.Enum().Values()
		allowedValues := violation.RuleValue.List()

		allowedNames := make([]string, allowedValues.Len())
		for i := range allowedValues.Len() {
			number := allowedValues.Get(i).Int()
			desc := allValues.ByNumber(protoreflect.EnumNumber(number))
			if desc == nil {
				allowedNames[i] = strconv.FormatInt(number, 10)
			} else {
				allowedNames[i] = string(desc.Name())
			}
		}

		slices.Sort(allowedNames)
		violation.Proto.Message = proto.String(fmt.Sprintf("must be one of [%s]", strings.Join(allowedNames, ", ")))
	}
}
```

This works nicely for enum fields in messages. However, it panics on enum-valued maps, because `violation.FieldDescriptor` is `nil`.

haines:
Fixes #322

### DIFF
--- a/error_utils.go
+++ b/error_utils.go
@@ -166,16 +166,22 @@ func FieldPathString(path *validate.FieldPath) string {
 	return result.String()
 }
 
-// markViolationForKey marks the provided error as being for a map key, by
-// setting the `for_key` flag on each violation within the validation error.
-func markViolationForKey(err error) {
+// markViolationForMapEntry marks the provided error as being for a map entry,
+// by setting the `for_key` flag on each violation within the validation error
+// as well as the field descriptor.
+func markViolationForMapEntry(err error, forKey bool, fieldDescriptor protoreflect.FieldDescriptor) {
 	if err == nil {
 		return
 	}
 	var valErr *ValidationError
 	if errors.As(err, &valErr) {
 		for _, violation := range valErr.Violations {
-			violation.Proto.SetForKey(true)
+			violation.Proto.SetForKey(forKey)
+			if forKey {
+				violation.FieldDescriptor = fieldDescriptor.MapKey()
+			} else {
+				violation.FieldDescriptor = fieldDescriptor.MapValue()
+			}
 		}
 	}
 }
--- a/map.go
+++ b/map.go
@@ -109,13 +109,14 @@ func (m *kvPairs) Evaluate(msg protoreflect.Message, val protoreflect.Value, cfg
 
 func (m *kvPairs) evalPairs(msg protoreflect.Message, key protoreflect.MapKey, value protoreflect.Value, cfg *validationConfig) (err error) {
 	evalErr := m.KeyRules.EvaluateField(msg, key.Value(), cfg, true)
-	markViolationForKey(evalErr)
+	markViolationForMapEntry(evalErr, true, m.Descriptor)
 	ok, err := mergeViolations(err, evalErr, cfg)
 	if !ok {
 		return err
 	}
 
 	evalErr = m.ValueRules.EvaluateField(msg, value, cfg, true)
+	markViolationForMapEntry(evalErr, false, m.Descriptor)
 	_, err = mergeViolations(err, evalErr, cfg)
 	return err
 }


rodaine:
Hey, @haines, thanks for the patch! Happy to make sure the violations have as much useful data as possible.

Couple of things:

When a map's values are messages, `ValueRules.EvaluateField` returns violations from the value-level rules as well as from any fields inside that value that also failed. The patch currently stamps every violation with the value's descriptor that were already correct. For example:

```protobuf
message Inner { 
  string name = 1 [buf.validate.field).string.min_len = 3]; 
}
message Outer { map<string, Inner> entries = 1; }
```

Validating `Outer{entries:{"k": {name: "ab"}}}` produces a violation with the field path `entries["k"].name` and (currently on main) the correct field descriptor of `Inner.name`. With this patch, the descriptor becomes `Outer.EntriesEntry.value` which is the map value's field (of type `Inner`).

`SetForKey(false)` on the value path ends up clobbering the violation as well. For example:

```protobuf
message Inner { 
  map<string, string> tags = 1 [(buf.validate.field).map.keys.string.min_len = 2]; 
}
message Outer { map<string, Inner> entries = 1; }
```

A key violation at `entries["k"].tags["x"]` is marked `for_key=true` by the inner map's key pass, and then the otuer map's value pass overwrites it to `false`. I think for the key path, you will still unconditionally stamp the violations since they're always scalars, but probably need a separate helper for the value path if those fields are unset.

Please also add regression tests to cover all these cases; don't need this subtly breaking in the future:

- a scalar value rule yields `MapValue` as the descriptor
- a key rule yields `MapKey`
- a nested field violation under a message-valued map keeps the inner field's descriptor with `for_key=false` (or to be consistent with the current impl, unset)
- a nested map-key violation keeps `for_key=true`

haines:
Hi @rodaine, thanks a lot for reviewing. Is there an existing test file where I should add these tests, or should I introduce a new one? It wasn't obvious to me where this functionality is currently being tested.

rodaine:
In `validator_test.go` would be perfect; I believe most of our regression tests are in there.

PR #325, opened 2026-06-15
Fix regression in Violation.RuleValue for native rules

### LINKED ISSUE #323: [BUG] `Violation.RuleValue` unreleased breaking change
## Description

Whilst attempting to create a PR to fix #322, I discovered that the `RuleValue` has changed since v1.2.0 and now returns an incorrect value for `enum.in`.

## Steps to Reproduce

https://github.com/haines/protovalidate-violation-rule-value

```proto
syntax = "proto3";

package example.v1;

import "buf/validate/validate.proto";

option go_package = "github.com/haines/protovalidate-violation-rule-value/gen/example/v1;examplev1";

enum Value {
  VALUE_UNSPECIFIED = 0;
  VALUE_A = 1;
  VALUE_B = 2;
}

message Example {
  Value value = 1 [(buf.validate.field).enum = {
    in: [
      1,
      2
    ]
  }];
}
```

```go
package main

import (
	"errors"
	"fmt"
	"log"

	"buf.build/go/protovalidate"
	examplev1 "github.com/haines/protovalidate-violation-rule-value/gen/example/v1"
)

func main() {
	err := protovalidate.Validate(&examplev1.Example{Value: examplev1.Value_VALUE_UNSPECIFIED})
	validationErr, ok := errors.AsType[*protovalidate.ValidationError](err)
	if !ok {
		log.Fatalln(err)
	}

	for _, violation := range validationErr.Violations {
		list := violation.RuleValue.List()
		for i := range list.Len() {
			fmt.Println(list.Get(i).Int())
		}
	}
}
```

With https://github.com/bufbuild/protovalidate-go/commit/8f3b41e4f38384e97070441c6dd3f76b6e419114:

```console
$ go run github.com/haines/protovalidate-violation-rule-value@v0.0.0-20260615193051-47e92f16d0df
1
2
```

With https://github.com/bufbuild/protovalidate-go/commit/e604f110506ca7d0c26c8b9a838d675895b4de51

```console
$ go run github.com/haines/protovalidate-violation-rule-value@v0.0.0-20260615193114-4eb14d8c0a92
panic: type mismatch: cannot convert int32 to list

goroutine 1 [running]:
google.golang.org/protobuf/reflect/protoreflect.Value.List({{}, 0x4aa600?, 0x0?, 0x4c93bc63d608?})
        /go/pkg/mod/google.golang.org/protobuf@v1.36.11/reflect/protoreflect/value_union.go:356 +0xd4
main.main()
        /go/pkg/mod/github.com/haines/protovalidate-violation-rule-value@v0.0.0-20260615193114-4eb14d8c0a92/main.go:20 +0xdc
```

## Expected Behavior

Per the documentation, `RuleValue` contains the value of the rule that specified the failed rule. For `enum.in`, this should be the list of values that are allowed.

## Actual Behavior

`RuleValue` appears to contain the `FieldValue` but with a different type (`Int` rather than `Enum`)?

## Screenshots/Logs

N/A

## Environment

- **Operating System**: N/A
- **Version**: N/A
- **Compiler/Toolchain**: N/A
- **Protobuf Compiler & Version**: buf v1.70.0, buf.build/protocolbuffers/go:v1.36.11
- **Protovalidate Version**: https://github.com/bufbuild/protovalidate-go/commit/e604f110506ca7d0c26c8b9a838d675895b4de51

## Additional Context

This appears to have been introduced in https://github.com/bufbuild/protovalidate-go/pull/316.

pkwarren:
The value passed to construct a Violation.RuleValue regressed as part of PR #316.

Update the native rules to fix inconsistencies between the CEL rules and the native rules. Add regression tests to verify the shape of the violation.

Fixes #323.

### DIFF
--- a/base.go
+++ b/base.go
@@ -139,6 +139,19 @@ func (b *base) newViolation(
 	}
 }
 
+func newListRuleValue[T any](
+	msg protoreflect.Message,
+	desc protoreflect.FieldDescriptor,
+	vals []T,
+	conv func(T) protoreflect.Value,
+) protoreflect.Value {
+	list := msg.NewField(desc).List()
+	for _, val := range vals {
+		list.Append(conv(val))
+	}
+	return protoreflect.ValueOfList(list)
+}
+
 func prefixRulePath(prefix *validate.FieldPath, suffix *validate.FieldPath) *validate.FieldPath {
 	if len(prefix.GetElements()) > 0 {
 		return validate.FieldPath_builder{
--- a/base.go
+++ b/base.go
@@ -139,13 +139,13 @@ func (b *base) newViolation(
 	}
 }
 
-func newListRuleValue[T any](
-	msg protoreflect.Message,
+func sliceToListValue[T any](
+	msg proto.Message,
 	desc protoreflect.FieldDescriptor,
 	vals []T,
 	conv func(T) protoreflect.Value,
 ) protoreflect.Value {
-	list := msg.NewField(desc).List()
+	list := msg.ProtoReflect().Mutable(desc).List()
 	for _, val := range vals {
 		list.Append(conv(val))
 	}
--- a/native_bytes.go
+++ b/native_bytes.go
@@ -132,22 +132,14 @@ func tryBuildNativeBytesRules(base base, rules *validate.BytesRules) evaluator {
 		hasRule = true
 	}
 
-	var (
-		inVals      [][]byte
-		inRuleValue protoreflect.Value
-	)
+	var inVals [][]byte
 	if inVals = rules.GetIn(); len(inVals) > 0 {
-		inRuleValue = newListRuleValue(rules.ProtoReflect(), bytesDescs.inSite.desc, inVals, protoreflect.ValueOfBytes)
 		rules.ProtoReflect().Clear(bytesDescs.inSite.desc)
 		hasRule = true
 	}
 
-	var (
-		notInVals      [][]byte
-		notInRuleValue protoreflect.Value
-	)
+	var notInVals [][]byte
 	if notInVals = rules.GetNotIn(); len(notInVals) > 0 {
-		notInRuleValue = newListRuleValue(rules.ProtoReflect(), bytesDescs.notInSite.desc, notInVals, protoreflect.ValueOfBytes)
 		rules.ProtoReflect().Clear(bytesDescs.notInSite.desc)
 		hasRule = true
 	}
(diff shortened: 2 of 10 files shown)


base.go
```diff
@@ -139,6 +139,19 @@ func (b *base) newViolation(
...
+	vals []T,
+	conv func(T) protoreflect.Value,
+) protoreflect.Value {
+	list := msg.NewField(desc).List()
+	for _, val := range vals {
+		list.Append(conv(val))
+	}
+	return protoreflect.ValueOfList(list)
+}
+
```
jonbodner-buf, line 154: we can make a protoreflect.Message as needed and not store the rule value if there are no errors :

```go
type ProtoReflector interface {
	ProtoReflect() protoreflect.Message
}

// sliceToListValue builds a detached protoreflect.List value for a repeated
// int32 enum rule field (enum.in / enum.not_in) from a Go slice. There is no
// API to wrap a raw slice as a protoreflect.List, so populate a fresh list off
// a new EnumRules message via the field descriptor. Called only when building
// a violation, so the cost is paid only on the failure path.
func sliceToListValue[T any](pr ProtoReflector, 
	desc protoreflect.FieldDescriptor, 
	vals []T, f func(T) protoreflect.Value) protoreflect.Value {
	list := pr.ProtoReflect().Mutable(desc).List()
	for _, v := range vals {
		list.Append(f(v))
	}
	return protoreflect.ValueOfList(list)
}
```

native_bytes.go
```diff
@@ -132,14 +132,22 @@ func tryBuildNativeBytesRules(base base, rules *validate.BytesRules) evaluator {
 		hasRule = true
 	}
 
-	var inVals [][]byte
+	var (
+		inVals      [][]byte
+		inRuleValue protoreflect.Value
+	)
 	if inVals = rules.GetIn(); len(inVals) > 0 {
+		inRuleValue = newListRuleValue(rules.ProtoReflect(), bytesDescs.inSite.desc, inVals, protoreflect.ValueOfBytes)
```
jonbodner-buf: do this in the error, not pre-emptively (repeat for the other cases too)

native_bytes.go
```diff
@@ -362,7 +374,7 @@ func (n nativeBytesEval) Evaluate(_ protoreflect.Message, val protoreflect.Value
 	if len(n.inVals) > 0 && !slices.ContainsFunc(n.inVals, func(v []byte) bool { return bytes.Equal(v, bytesVal) }) {
 		violations = append(violations, n.newViolation(bytesDescs.inSite,
 			"bytes.in", "must be in list "+formatBytesList(n.inVals),
-			val, protoreflect.ValueOfBytes(bytesVal)))
+			val, n.inRuleValue))
```
jonbodner-buf: change to:

```go
		violations = append(violations, n.newViolation(bytesDescs.inSite,
			"bytes.in", "must be in list "+formatBytesList(n.inVals),
			val, sliceToListValue(&validate.BytesRules{}, bytesDescs.inSite.desc, n.inVals, protoreflect.ValueOfBytes)))
```

jonbodner-buf: and so on for the other types.

jonbodner-buf (changes requested):

jonbodner-buf (approved):

PR #327, opened 2026-07-02
Add ValidateContext for cancellation and timeout support

tonobo:
Add a context-aware ValidateContext(ctx, msg, options...) function alongside Validate() to let callers constrain lengthy or complex validations (e.g., CEL rule evaluation over large collections) using context cancellation or deadlines.

Content:
- Add a ContextValidator interface (Validator plus ValidateContext). New now returns ContextValidator, so New() callers get ValidateContext directly while still satisfying Validator. This keeps the change backward compatible: the Validator interface is unchanged, so external implementations of Validator keep compiling.
- Add the package-level ValidateContext helper and implement ValidateContext on the global validator. Validate() delegates to ValidateContext(context.Background(), ...), keeping a single code path.
- Thread context through every evaluator via parallel *Context methods; the non-context methods delegate with context.Background(), so behavior is unchanged on the Validate() path.
- Evaluate CEL through cel-go's ContextEval and enable InterruptCheckFrequency so a single long-running expression can be interrupted mid-evaluation.
- Check ctx.Err() at message entry and in the repeated/map iteration loops so native validation of large messages also honors cancellation.
- Return the raw context error (context.Canceled / DeadlineExceeded), never wrapped, so callers can match it with errors.Is.

### DIFF
--- a/cel/library.go
+++ b/cel/library.go
@@ -40,8 +40,10 @@ import (
 //
 // Using this function, you can create a CEL environment that is identical to
 // the one used to evaluate protovalidate CEL expressions.
-func NewLibrary() cel.Library {
-	return &library{
+// Options may be passed to configure the library; see [LibraryOption].
+func NewLibrary(options ...LibraryOption) cel.Library {
+	lib := &library{
+		interruptCheckFrequency: DefaultInterruptCheckFrequency,
 		uniqueScalarPool: sync.Pool{New: func() any {
 			return map[ref.Val]struct{}{}
 		}},
@@ -49,6 +51,37 @@ func NewLibrary() cel.Library {
 			return map[string]struct{}{}
 		}},
 	}
+	for _, option := range options {
+		option(lib)
+	}
+	return lib
+}
+
+// DefaultInterruptCheckFrequency is the default value for
+// [WithInterruptCheckFrequency].
+//
+// cel-go checks for context cancellation only inside comprehension loops
+// (macros such as all, exists, map and filter), once per iteration. The
+// frequency selects how many of those iterations pass between successive polls
+// of ctx.Done(): the per-iteration bookkeeping happens regardless, and the
+// frequency only gates the (very cheap) non-blocking channel poll. Raising it
+// therefore buys no measurable throughput, while making cancellation latency
+// proportionally worse. 1 is the responsive default.
+const DefaultInterruptCheckFrequency = 1
+
+// LibraryOption configures the CEL library returned by [NewLibrary].
+type LibraryOption func(*library)
+
+// WithInterruptCheckFrequency sets how many comprehension iterations elapse
+// between checks for context cancellation during CEL evaluation.
+//
+// A value of 0 disables interrupt checking entirely: CEL expressions then run
+// to completion and cannot be cancelled mid-evaluation.
+//
+// Note that expressions containing no comprehension are never interruptible,
+// whatever the frequency, because cel-go only checks inside comprehension loops.
+func WithInterruptCheckFrequency(frequency uint) LibraryOption {
+	return func(lib *library) { lib.interruptCheckFrequency = frequency }
 }
 
 // library is the collection of functions and settings required by protovalidate
@@ -59,8 +92,9 @@ func NewLibrary() cel.Library {
 // All implementations of protovalidate MUST implement these functions and
 // should avoid exposing additional functions as they will not be portable.
 type library struct {
-	uniqueScalarPool sync.Pool
-	uniqueBytesPool  sync.Pool
+	uniqueScalarPool        sync.Pool
+	uniqueBytesPool         sync.Pool
+	interruptCheckFrequency uint
 }
 
 func (l *library) CompileOptions() []cel.EnvOption { //nolint:funlen,gocyclo
@@ -378,11 +412,15 @@ func (l *library) CompileOptions() []cel.EnvOption { //nolint:funlen,gocyclo
 }
 
 func (l *library) ProgramOptions() []cel.ProgramOption {
-	return []cel.ProgramOption{
+	opts := []cel.ProgramOption{
 		cel.EvalOptions(
 			cel.OptOptimize,
 		),
 	}
+	if l.interruptCheckFrequency > 0 {
+		opts = append(opts, cel.InterruptCheckFrequency(l.interruptCheckFrequency))
+	}
+	return opts
 }
 
 func (l *library) uniqueMemberOverload(itemType *cel.Type, overload func(lister traits.Lister) ref.Val) cel.FunctionOpt {
--- a/evaluator.go
+++ b/evaluator.go
@@ -15,6 +15,8 @@
 package protovalidate
 
 import (
+	"context"
+
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
@@ -36,6 +38,9 @@ type evaluator interface {
 	//       build. This error is not recoverable.
 	//
 	Evaluate(msg protoreflect.Message, val protoreflect.Value, cfg *validationConfig) error
+
+	// EvaluateContext is like Evaluate, but honors context cancellation.
+	EvaluateContext(ctx context.Context, msg protoreflect.Message, val protoreflect.Value, cfg *validationConfig) error
 }
 
 // messageEvaluator is essentially the same as evaluator, but specialized for
@@ -46,6 +51,10 @@ type messageEvaluator interface {
 	// EvaluateMessage checks that the provided msg is valid. See
 	// evaluator.Evaluate for behavior
 	EvaluateMessage(msg protoreflect.Message, cfg *validationConfig) error
+
+	// EvaluateMessageContext is like EvaluateMessage, but honors context
+	// cancellation.
+	EvaluateMessageContext(ctx context.Context, msg protoreflect.Message, cfg *validationConfig) error
 }
 
 // evaluators are a set of evaluator applied together to a value. Evaluation
@@ -53,10 +62,14 @@ type messageEvaluator interface {
 // true or a different error is returned.
 type evaluators []evaluator
 
-func (e evaluators) Evaluate(msg protoreflect.Message, val protoreflect.Value, cfg *validationConfig) (err error) {
+func (e evaluators) Evaluate(msg protoreflect.Message, val protoreflect.Value, cfg *validationConfig) error {
+	return e.EvaluateContext(context.Background(), msg, val, cfg)
+}
+
+func (e evaluators) EvaluateContext(ctx context.Context, msg protoreflect.Message, val protoreflect.Value, cfg *validationConfig) (err error) {
 	var ok bool
 	for _, eval := range e {
-		evalErr := eval.Evaluate(msg, val, cfg)
+		evalErr := eval.EvaluateContext(ctx, msg, val, cfg)
 		if ok, err = mergeViolations(err, evalErr, cfg); !ok {
 			return err
 		}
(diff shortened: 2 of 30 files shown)


tonobo:
@pkwarren Would you like to give it a shot? :smile:

rodaine:
Hi, @tonobo, thanks for the patch! I'd like to dig into your use-case here, could you share some details on the size of the message you're validating where cancellation is relevant? What type of validation rules are you using that are this CPU intensive?

It's cool to see that cel-go added support for context-aware execution, but I'm wary of foisting this behavior onto all users of this library when most messages are evaluated on the order of a microsecond. I'd love it if you could add benchmarks to show that this does not regress the default usage.

tonobo:
Hey, great to hear someone is interested in this contribution!

First, the use case: imagine a global repository managing all contracts, so
besides the actual proto definitions there is also type validation in place.
As time goes on, types get more nested and more complicated. CEL rules start
iterating over objects, and one thing leads to another: inefficient rules get
introduced. Nothing actually prevents you from writing nested loops over
maps :D

To make this concrete, here is an example. A message holds a list of named sets plus several rule lists that reference
those sets by key, and cross-field CEL rules enforce referential integrity:

```protobuf
    message Firewall {
      option (buf.validate.message).cel = {
        id: "ipset_keys_unique"
        message: "ipset keys must be unique within this firewall"
        expression: "this.ipsets.map(s, string(s.key.type) + ':' + s.key.name).unique()"
      };

      // one of these exists per rule list (ingress/egress, several stages),
      // so this pattern repeats 4+ times on the same message
      option (buf.validate.message).cel = {
        id: "ingress_ipsets_exist"
        message: "all ingress rules must only reference ipsets which exist within this firewall"
        expression:
          "this.ingress.rules.all(rule, rule.remote_ipsets.all(key,"
          "this.ipsets.exists(ipset, ipset.key.type == key.type && ipset.key.name == key.name)))"
      };

      // ...
    }
```

Each of those rules is O(rules * refs_per_rule * ipsets), so validation cost
grows roughly cubically with message size, and there are several such rules
per message. Perfectly reasonable contracts to enforce, but on large messages
the evaluation time explodes.

CEL provides different options to tackle this. On one end, you can set
program-based cost limits. I considered that as well, but program compilation
happens at a different level, so it would have to be an option on the
initializer, not a per-validation option. That's why we decided to go the
other way around and address the problem from the other end: context
cancellation. AFAIR cel-go internally uses context.Background() anyway, the
main adjustment is the InterruptCheckFrequency() we attach to the evaluation.
I haven't fully dug into the cel-go internals, but it probably just adds a
periodic check and might cost a few extra cycles. My guess is the overhead is
minimal.

In conclusion: protovalidate is used to enforce contracts between parties,
and context cancellation brings more trust into "anonymous rules". Assume a
bad case: a message in the queue costs you 10s to validate, your workers are
busy validating protos all the time, and suddenly quadratic (or worse) CEL
expression costs are a real problem.

Regarding the benchmarks: fair point! I'll attach a benchmark test comparing
the default branch against this one.

tonobo:
*  Validate no longer pays for context support: the validator now detects whether the context can be cancelled at all (ctx.Done() != nil). If
  not, it skips every ctx.Err() check and uses plain Eval instead of ContextEval, so the non-context path costs the same as before this PR.
  Benchmarks added (BenchmarkContextOverhead).

* New BenchCrossReference test message with an expensive cross-referencing CEL rule, used to prove a deadline actually interrupts a
  long-running expression.

tonobo:
@rodaine Is there anything else I can do for you? It would be great if we could make some progress here :D Sorry for bugging you. :see_no_evil:

cel/library.go
```diff
@@ -40,15 +40,48 @@ import (
...
+// [WithInterruptCheckFrequency].
+//
+// cel-go checks for context cancellation only inside comprehension loops
+// (macros such as all, exists, map and filter), once per iteration. The
+// frequency selects how many of those iterations pass between successive polls
+// of ctx.Done(): the per-iteration bookkeeping happens regardless, and the
+// frequency only gates the (very cheap) non-blocking channel poll. Raising it
+// therefore buys no measurable throughput, while making cancellation latency
+// proportionally worse. 1 is the responsive default.
+const DefaultInterruptCheckFrequency = 1
```
pkwarren, line 70: We should remove this constant and interrupt checks should be disabled by default (matching CEL's defaults). To this point no users of protovalidate-go have requested this feature and shouldn't require this additional overhead by default. We can always re-evaluate the default later.

evaluator.go
```diff
@@ -36,6 +38,9 @@ type evaluator interface {
 	//       build. This error is not recoverable.
 	//
 	Evaluate(msg protoreflect.Message, val protoreflect.Value, cfg *validationConfig) error
+
+	// EvaluateContext is like Evaluate, but honors context cancellation.
+	EvaluateContext(ctx context.Context, msg protoreflect.Message, val protoreflect.Value, cfg *validationConfig) error
```
pkwarren, line 43: Instead of overloads on the evaluator (internal interface) all the way down we should just update the signature for `Evaluate` to take a `ctx context.Context`.

evaluator.go
```diff
@@ -46,17 +51,25 @@ type messageEvaluator interface {
 	// EvaluateMessage checks that the provided msg is valid. See
 	// evaluator.Evaluate for behavior
 	EvaluateMessage(msg protoreflect.Message, cfg *validationConfig) error
+
+	// EvaluateMessageContext is like EvaluateMessage, but honors context
+	// cancellation.
+	EvaluateMessageContext(ctx context.Context, msg protoreflect.Message, cfg *validationConfig) error
```
pkwarren, line 57: Same here for `EvaluateMessage`.

map.go
```diff
@@ -107,15 +118,15 @@ func (m *kvPairs) Evaluate(msg protoreflect.Message, val protoreflect.Value, cfg
 	return err
 }
 
-func (m *kvPairs) evalPairs(msg protoreflect.Message, key protoreflect.MapKey, value protoreflect.Value, cfg *validationConfig) (err error) {
-	evalErr := m.KeyRules.EvaluateField(msg, key.Value(), cfg, true)
+func (m *kvPairs) evalPairsContext(ctx context.Context, msg protoreflect.Message, key protoreflect.MapKey, value protoreflect.Value, cfg *validationConfig) (err error) {
```
pkwarren, line 121: Just change the existing method to take a context as the first argument instead of changing every method to add `Context` suffix.

map.go
```diff
@@ -66,10 +67,20 @@ func newKVPairs(valEval *value) *kvPairs {
...
+func (m *kvPairs) Evaluate(msg protoreflect.Message, val protoreflect.Value, cfg *validationConfig) error {
+	return m.EvaluateContext(context.Background(), msg, val, cfg)
+}
+
+func (m *kvPairs) EvaluateContext(ctx context.Context, msg protoreflect.Message, val protoreflect.Value, cfg *validationConfig) (err error) {
 	var ok bool
 	val.Map().Range(func(key protoreflect.MapKey, value protoreflect.Value) bool {
-		evalErr := m.evalPairs(msg, key, value, cfg)
+		if cfg.cancellable {
+			if ctxErr := ctx.Err(); ctxErr != nil {
```
pkwarren, line 78: We should update these checks to use `context.Cause(ctx)` instead of `ctx.Err()`.

program.go
```diff
@@ -41,16 +45,31 @@ func (s programSet) Eval(
...
+	ctx context.Context,
+	val protoreflect.Value,
+	fieldDesc protoreflect.FieldDescriptor,
+	cfg *validationConfig,
 ) error {
 	if len(s.programs) == 0 {
 		return nil
 	}
+	if cfg.cancellable {
+		if err := ctx.Err(); err != nil {
```
pkwarren, line 63: Should use `context.Cause(ctx)` here.

program.go
```diff
@@ -91,17 +110,41 @@ type compiledProgram struct {
...
+		// the activation on every evaluation just to poll a nil channel. This
+		// keeps the Validate path free of any context-related overhead.
+		value, _, err = expr.Program.Eval(activation)
+	}
 	if err != nil {
+		// Only cel-go's interrupt error signals cancellation; any other failure
+		// is a genuine evaluation error and must keep its RuntimeError identity
+		// (with the rule id) even if the context has expired in the meantime.
+		if errors.Is(err, interpreter.InterruptError{}) {
+			if ctxErr := ctx.Err(); ctxErr != nil {
```
pkwarren, line 144: Use `context.Cause(ctx)` here.

program_test.go
```diff
@@ -287,6 +287,62 @@ func TestSet_BindThis(t *testing.T) {
 	}
 }
 
+func TestCompiled_ContextCancelled(t *testing.T) {
+	t.Parallel()
+	ctx, cancel := context.WithCancel(context.Background())
```
pkwarren, line 292:
```suggestion
	ctx, cancel := context.WithCancel(t.Context())
```

validator.go
```diff
@@ -52,13 +53,43 @@ type Validator interface {
...
-// individual ValidatorOption for how they impact the fallibility of New.
-func New(options ...ValidatorOption) (Validator, error) {
+// ContextValidator is a [Validator] that additionally supports context-aware
+// validation. The value returned by [New] and the [GlobalValidator] both
+// implement it. Callers that hold only a [Validator] can type-assert it to
+// ContextValidator, or use the package-level [ValidateContext] function.
+//
+// ContextValidator is a separate interface (rather than a method on Validator)
+// to preserve backward compatibility for external implementations of Validator.
+type ContextValidator interface {
```
pkwarren, line 63: Instead of introducing a new `ContextValidator` and changing the primary API for users, we should add a `ValidationOption` named `WithContext` which takes a context to bound the validate call. The default value would be `context.Background()` if unset.

The rest of the changes (checking cancelable, passing down a context to evaluators) would largely stay the same.

pkwarren (commented):
Thank you for adding some benchmarks. It would be good to gather `benchstat` output from before/after this change to ensure no regressions.

Overall the changes look good but this should be opt-in behavior for users with expensive validation constraints.

validator.go
tonobo: I refactored it to the WithContext(ctx) option as suggested. One observation from benchmarking: passing a context as an option is a bit uncommon as an API, and the option mechanism itself is what raises the cost, not the context support. On a trivial scalar message, Validate(msg) is ~67ns/0 allocs while Validate(msg, WithContext(ctx)) is ~130ns/2 allocs - but ~90% of that is generic ValidationOption plumbing (interface boxing and the variadic slice at the call site, plus the config copy; WithFilter pays the same today). The cancellation checks and cel-go's ContextEval add only ~6ns; a never-cancellable context skips them entirely. The cost is constant per call, so on realistic messages it disappears (ComplexSchema +2.4%). I also added a sync.Pool for the per-call config copy in a separate commit, which removes the largest of the three allocations. If the per-call overhead ever matters, the alloc-free design would be a dedicated method taking the context - but I understand the preference for keeping the Validator interface unchanged, and the numbers above are fine for the intended use.


```
goos: linux / goarch: amd64 / Ryzen 7 PRO 7840U, n=6

sec/op                     before (main)   after (branch)         after + WithContext (cancellable)
Scalar                     66.99n ±1%      66.72n   ~             133.10n   +98.69%
Repeated/Scalar            95.39n ±3%      95.06n   ~             171.80n   +80.10%
Repeated/Message           274.7n ±3%      279.4n   ~             432.5n    +57.47%
Repeated/Unique/Scalar     294.3n ±1%      300.4n   +2.06%        389.5n    +32.33%
Repeated/Unique/Bytes      474.0n ±4%      472.1n   ~             612.9n    +29.30%
Map                        103.8n ±2%      104.9n   ~             173.3n    +67.00%
ComplexSchema              14.89u ±2%      14.64u   -1.66%        15.24u    +2.39%
MultiRuleError             812.6n ±5%      830.3n   ~             890.2n    +9.55%
MultiRuleNoError           66.73n ±1%      67.11n   ~             129.65n   +94.29%
geomean                    301.4n          303.0n   +0.53%        447.5n    +48.49%

B/op        after == before everywhere; WithContext: +32 B/op flat
allocs/op   after == before everywhere; WithContext: +2 allocs/op flat
```

PR #332, opened 2026-07-30
Exclude the trailing dot from the hostname 253-character limit

LordAizen1:
`IsHostname` checked the length before stripping the trailing dot, so a valid 253-character name with a trailing dot (254 bytes) got rejected. The 253 limit excludes that optional dot, so it should pass. Fix is to strip the dot first, then check the length.

This is the runtime side of bufbuild/protovalidate#512. The matching conformance case `valid/name_253_characters_plus_trailing_dot` landed in bufbuild/protovalidate#513, and @rodaine suggested following up with the implementation patches.

Added boundary cases to `TestIsHostname` (253 + dot valid, 254 and 254 + dot still invalid).

### DIFF
--- a/internal/rules/rules.go
+++ b/internal/rules/rules.go
@@ -130,14 +130,16 @@ func IsIPPrefix(
 //   - The name can have a trailing dot, for example "foo.example.com.".
 //   - The name can be 253 characters at most, excluding the optional trailing dot.
 func IsHostname(val string) bool {
-	if len(val) > 253 {
-		return false
-	}
 	var str string
 	if strings.HasSuffix(val, ".") {
 		str = val[0 : len(val)-1]
 	} else {
 		str = val
+	}
+	// The 253-character limit excludes the optional trailing dot, so measure
+	// after stripping it.
+	if len(str) > 253 {
+		return false
 	}
 
 	allDigits := false
--- a/internal/rules/rules_test.go
+++ b/internal/rules/rules_test.go
@@ -15,6 +15,7 @@
 package rules
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


internal/rules/rules.go
```diff
@@ -130,15 +130,17 @@ func IsIPPrefix(
...
-		return false
-	}
 	var str string
 	if strings.HasSuffix(val, ".") {
 		str = val[0 : len(val)-1]
 	} else {
 		str = val
 	}
+	// The 253-character limit excludes the optional trailing dot (RFC 3339),
+	// so measure after stripping it.
```
timostamm:
```suggestion
	// The 253-character limit excludes the optional trailing dot, so measure
	// after stripping it.
```

RFC 3339 is something else :)

RFC 1035 doesn't specify the limits for the text form, only for the wire format. On the wire, names are always absolute, but in text they can be relative. The trailing dot makes a name absolute, but it stands for the root label, which the 255-octet limit already counts.

internal/rules/rules_test.go
```diff
@@ -30,6 +31,16 @@ func TestIsHostname(t *testing.T) {
 	require.True(t, IsHostname("foo.example.com"))
 	require.True(t, IsHostname("A.ISI.EDU"))
 	require.False(t, IsHostname("İ"))
+
+	// The 253-character limit excludes the optional trailing dot (RFC 3339).
```
timostamm:
```suggestion
	// The 253-character limit excludes the optional trailing dot.
```

timostamm (commented):
I've run the new conformance test locally, and the change passes. Please drop the RFC mention from the comments, see below, then this should be ready to merge.

LordAizen1:
Good catch, thanks for the explanation. Dropped the RFC mention from both comments (and the commit message and description) to match your suggestions.

timostamm (approved):
Thanks for the fix!

PR #333, opened 2026-08-05
Use CEL PartialVars during AST pruning

TristonianJones:
A change in cel-go v0.30.0 revealed a latent bug in how residual expressions were computed with protovalidate-go, as `error` values terminate function call argument evaluation. The fix marks `this` as unknown and calls CEL with a `PartialVars` argument to clearly denote the desired, non-terminating, evaluation behavior.

Metric | Before Patch | After Patch | Change / Improvement
-- | -- | -- | --
Allocations | 241 allocs/op | 5 allocs/op | ~98% reduction (48x fewer allocs)
Memory Allocated | 44,059 B/op (~44 KB) | 96 B/op | ~99.8% reduction (>450x less memory)
Execution Time | 16,856 ns/op (~16.8 µs) | 404.1 ns/op (~0.4 µs) | ~42x faster
Throughput | 69,870 ops | 2,931,400 ops | ~42x throughput

### DIFF
--- a/ast.go
+++ b/ast.go
@@ -61,7 +61,12 @@ func (set astSet) ReduceResiduals(rules protoreflect.Message, opts ...cel.Progra
 			residuals = append(residuals, ast)
 			continue
 		}
-		val, details, _ := program.Program.Eval(activation)
+		partialAct, err := cel.PartialVars(activation, cel.AttributePattern("this"))
+		if err != nil {
+			residuals = append(residuals, ast)
+			continue
+		}
+		val, details, _ := program.Program.Eval(partialAct)
 		if val != nil {
 			switch value := val.Value().(type) {
 			case bool:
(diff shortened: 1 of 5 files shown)


TristonianJones:
@miparnisari FYI, as this addresses https://github.com/cel-expr/cel-go/issues/1403

miparnisari:
Thank you @TristonianJones! Is the `after patch` column a return to pre-0.30.0 behavior, or is that _plus_ an actual improvement?

rodaine:
Oh sweet! Thanks for the patch, @TristonianJones. Looks like there are some formatting issues, but otherwise looks good.

TristonianJones:
Hi @rodaine, I believe I updated the formatting issue. Would you mind reapproving the workflow?

The issue in question is a regression from cel-go v0.29.2. As @miparnisari the 'before' patch here is how things regressed, and the 'after' is restoring it to its previous performance.

rodaine (approved):

rodaine:
Thanks again!

tstirrat15:
Any sense of when this will go out? We're keen to see it.

pkwarren:
> Any sense of when this will go out? We're keen to see it.

Version 1.3.0 is now available which includes this fix.

PR #337, opened 2026-08-24
Update deps and minimum Go version to 1.26

### LINKED ISSUE #335: [BUG] cel-go dependency relocated
## Description

As per https://github.com/cel-expr/cel-go/releases/tag/v0.32.0, cel-go has been relocated.

```diff
- module github.com/google/cel-go
+ module cel.dev/cel-go
```

rodaine:
Fixes #335 as well.


go.mod
```diff
@@ -1,25 +1,25 @@
 module buf.build/go/protovalidate
 
-go 1.25.0
+go 1.26.5
```
pkwarren: Do we require a patch version here from a dependency or can we leave this as `go 1.26.0` or `go 1.26`?

pkwarren (approved):

go.mod
jonbodner-buf: 1.26.7 is the latest patch version. Should we specify it?

rodaine: I tried to set it to either and I'm not why but I was getting build errors. Let me try again and see if I was fighting Goland or something...

rodaine: Yeah I think that's the version I have on my machine rn. See @pkwarren's comment.

pkwarren (approved):

go.mod
jonbodner-buf: https://blog.howardjohn.info/posts/go-mod-version/ .If we follow this advice (I'm not sure it's correct), we should set the go version to 1.26.0.

rodaine: Yep that's what I did

PR #339, opened 2026-09-04
Use native rules for repeated item rules

sudorandom:
`repeated.items` rules always fell back to CEL, even with native rules enabled. The native evaluators already work per-element, so letting the dispatch run for items is enough: a validate of a 1,000-item field with `items.int32.gt` drops from 96 µs to 20 µs, and the repo benchmarks improve 1.5–4.7%.

```proto
message Review {
  repeated int32 ratings = 1 [(buf.validate.field).repeated = {
    max_items: 100 // native
    items: {
      int32: {
        gte: 1 // Executes in CEL for each element
        lte: 5 // Executes in CEL for each element
      }
    }
  }];
}
```

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -474,24 +474,24 @@ func (bldr *builder) processStandardRules(
 	// reused (happens in some test cases, could happen in production code with dynamic messages).
 	rules = proto.CloneOf[*validate.FieldRules](rules)
 
-	// put behind a feature flag to allow for testing.
-	// it's easier to follow like this, don't break it up
-	//nolint:nestif
+	// Try native Go evaluators for known simple rules before falling back to
+	// CEL. put behind a feature flag to allow for testing.
 	if !bldr.disableNativeRules {
-		// Try native Go evaluators for repeated list-level rules (min_items, max_items, unique).
-		if fdesc.IsList() && valEval.NestedRule == nil {
+		switch {
+		case fdesc.IsList() && valEval.NestedRule == nil:
+			// List-level rules (min_items, max_items, unique).
 			if native := tryNativeRepeatedRules(newBase(valEval), rules.GetRepeated()); native != nil {
 				valEval.Append(native)
 			}
-		}
-		// Try native Go evaluators for map-level rules (min_pairs, max_pairs).
-		if fdesc.IsMap() && valEval.NestedRule == nil {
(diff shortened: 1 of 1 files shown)


rodaine:
Do map keys/values also need this treatment?

sudorandom:
@rodaine good question. The answer is no. Map keys/values are already using native rules as intended.

For maps, when the builder recurses into map rules, it uses the entry's key/value field descriptors, which are plain singular fields accessed at `m.MapKey()` and `m.MapValue()`, so the original `!IsMap() && !IsList()` condition worked fine for them.

Repeated items are a bit different. A repeated field (aka list) has no separate descriptor for its elements, so the builder re-uses the repeated field's descriptor, where `IsList()` is still true for both the list and its elements. That's why the condition needed the `valEval.NestedRule != nil` part; when that's set, we're compiling rules for an individual item, not the list itself.

rodaine (approved):

builder.go
```diff
@@ -491,7 +491,7 @@ func (bldr *builder) processStandardRules(
 			}
 		}
 		// Try native Go evaluators for known simple rules before falling back to CEL.
-		if !fdesc.IsMap() && !fdesc.IsList() {
+		if !fdesc.IsMap() && (!fdesc.IsList() || valEval.NestedRule != nil) {
```
jonbodner-buf: There's already a case on lines 481-485 to handle repeated rules when there are no nested rules. Should the fix be to eliminate the check for nested rules in the earlier code? I haven't looked at this code in a while, I might be missing a detail.

sudorandom: Not quite. Those lines are to figure out when to evaluate rules related to list-level rules. However, you do have a point that this code is confusingly organized here. I switched this to use a much more understandable switch statement.

To be clear, the logic error is that this section REQUIRES NestedRules to be set,  aka `!= nil`, so it's different than the case above.

jonbodner-buf (approved):

builder.go
jonbodner-buf: The switch isn't quite the same logic as before. The default case used to only process scalar values. Now it will process fdesc.IsList() and feds.IsMap() when valEval.NestedRule != nil . If that works correctly, great.

jonbodner-buf (approved):

